PR #41, opened 2015-02-11
Fix remove gen policy

dohse:


GayathriKaliyamoorthy:
The current code has these changes.

PR #56, opened 2015-05-06
Update aerospike-client-c.sh to support Linux Mint 17

jiminikiz:
I have seen plenty of discussion about adding support to LinuxMint17. The solution has also been posted. After adding this simple check, it works flawless on LinuxMint17.

Please consider adding this pull request so that when developers try installing aerospoike via NPM the process goes smoothly.

### DIFF
--- a/scripts/aerospike-client-c.sh
+++ b/scripts/aerospike-client-c.sh
@@ -61,7 +61,7 @@ detect_linux()
         return 0
         ;;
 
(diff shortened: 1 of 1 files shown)


jiminikiz:
I call bad test case.

GayathriKaliyamoorthy:
Updated manually as the original repo no longer exists.

Thanks

jiminikiz:
Thank you!

PR #69, opened 2015-07-02
Update nan & link includes folder

dohse:


GayathriKaliyamoorthy:
Cherry picked the commit which updates nan to 1.8.4.

dohse:
The second commit avoids copying `/usr/include` during build of the module. That folder is currently 150 MB on my machine.

PR #81, opened 2015-07-31
ran linter on js libs, added debug module

fed135:
Linter options are:
```
{
	// exposed jsbeautifier options
	"indent-size": 2,
	"indent_with_tabs": false,
	"preserve_newlines": true,
	"max_preserve_newlines": 4,
	"space_in_paren": false,
	"jslint_happy": false,
	"brace_style": "collapse",
	"keep_array_indentation": true,
	"keep_function_indentation": true,
	"eval_code": false,
	"unescape_strings": false,
	"break_chained_methods": false,
	"e4x": false,
	"wrap_line_length": 0,

	// jsformat options
	"format_on_save": false,
	"jsbeautifyrc_files": false
}

### DIFF
--- a/lib/aerospike.js
+++ b/lib/aerospike.js
@@ -1,7 +1,8 @@
-var as = require('../build/Release/aerospike.node')
+var as = require('../build/Release/aerospike.node');
 var LargeList = require('./llist.js');
 var stream = require('stream');
 var inherits = require('util').inherits;
+var stdErr = require('debug')('aerospike:error');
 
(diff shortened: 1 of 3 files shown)


GayathriKaliyamoorthy:
I have cherry picked this commit and it's available from release 1.0.49. I have removed the dependency on debug module for now. Once we finalize on the logger (winston/debug) as mentioned in the other thread #82, we will include the corresponding one.

Thanks.

PR #96, opened 2015-11-24
Preventing leakage to global scope + allowing LargeList instances

mrbar42:
##### Change 1:
top level explicit variable declarations are missing and strict mode is not enabled.
variables are leaking passively to global scope which can cause collision with other libraries and user modules.

strict mode should also be enabled in my opinion.

list of restrained variables: `LargeList, execute, Info,query, createIntegerIndex, createStringIndex, createGeo2DSphereIndex, parseOperateArgs, add, append, prepend ,`

##### Change 2:
Allowing instances in LargeLists.

in the current formation this is not possible:
``` javascript
var list1 = client.LargeList({ns: 'test', set: 'test', key: 'my_key1'}, 'ldt_bin');
var list2 = client.LargeList({ns: 'test', set: 'test', key: 'my_key2'}, 'ldt_bin');
```
Because LargeList stores arguments on itself (shared among all calls).
Generally speaking this is an anti pattern.
``` javascript
var LargeList = function (key, binName, writePolicy, createModule) {
    LargeList.key = key;
    return LargeList
}
```

Offered change:
``` javascript
var LargeList = function (key, binName, writePolicy, createModule) {
    var instance = {};
    instance.key = key;
    return instance
}
```
this will also encourage users to reuse object.
The structure is still wasteful because all methods are attached to each instance instead of using prototype or other method sharing techniques.

### DIFF
--- a/lib/aerospike.js
+++ b/lib/aerospike.js
@@ -25,7 +25,7 @@ RecordStream.prototype._read = function() {};
 // to node.js layer.
 // On error, emits 'error' event.
 // When all the results are consumed it emits 'end' event.
-execute = function() {
+var execute = function() {
 
 	var rs = new RecordStream();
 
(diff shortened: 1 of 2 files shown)


GayathriKaliyamoorthy:
Hi @mrbar42,

Thanks for the PR. I am merging this with our internal repo. This will be part of our next official release.

Issue #105, opened 2016-01-26
Discussion: error first callbacks, API changes


eljefedelrodeodeljefe:
As per discussion on #11 and short email correspondence with Alvin R. of Aerospike, I want to propose some API changes. Those might be friendlier to the `node` developer as it adopts `error-first`callbacks being `null` on success and generally rather resemble other DB interfaces.

**Motivation:**
The DB is awesome, but the client's API is hard to teach and kind of verbose. 

Since porting the tests will take a while I wanted to have the discussion and proposal before the effective PR. However afterwards the API should look like the below, with the library changes below that.

```js
'use strict'
const Aerospike = require('./lib/aerospike')

let config = {
  hosts: [
    { addr: '127.0.0.1', port: 3000 }
  ]
}

Aerospike.connect(config, function (err) {
  if (err) {
    console.log(err);
  }

  var key = Aerospike.key('test','demo','foo');
  console.log(key);

  var bins = { i : 123, s : "str"};
  var metadata = {ttl : 10000, gen: 1};

  Aerospike.put(key, bins, metadata, function(err, key) {
    if (err) {
      console.log('put error:', err);
    }

    Aerospike.get(key, function(err, rec, meta) {
      if (err) {
        console.log('get error:', err);
      }
      console.log(rec, meta);
    });
  });
})
```

It's fairly easy to port this with the current `cpp` layer below.

```js
var as = require('../build/Release/aerospike.node')

function Aerospike () {
  this.key = as.key,
  this.status = as.status,
  this.status_ok = this.status.AEROSPIKE_OK,
  this.policy = as.policy,
  this.language = as.language,
  this.log = as.log,
  this.scanPriority = as.scanPriority,
  this.predicates = as.predicates,
  this.indexType = as.indexType,
  this.scanStatus = as.scanStatus,
  this.Double = as.Double

  this._currentClient = {}
  this.proto = {}
}

Aerospike.prototype._client = function client (config) {
  let client = as.client(config)
  if (client === null) {
    throw new Error('Client object creation failed - null value returned')
  }

  this.proto = Object.getPrototypeOf(client)

  // TODO: implement subclases
  // if (!proto.createQuery) {
  //   proto.createQuery = proto.query
  //   proto.query = query
  // }

  // if (!proto.createIntegerIndex) {
  //   proto.createIntegerIndex = createIntegerIndex
  // }
  //
  // if (!proto.createStringIndex) {
[...]

PR #106, opened 2016-01-26
api: error-first callbacks

eljefedelrodeodeljefe:
as per discussion on #105 

This is probably a `semver`major change, since it's not only introducing a `js` class, to have error first callbacks, but also changing current error handling behaviour.

Todos:

- [x] client API in js
- [x] implement subclasses
- [x] ~~remove~~ re-enable boilerplate in tests, e.g. `before`, `after`; remove ~~`const status_ok`~~, all done for implemented tests. Rest will be changed along the way.

- [x] port tests
   - [x] get
   - [x] put
       - [x] `Doubles` not working
   - [x] select
   - [x] batch_exists
   - [x] batch_get
   - [x] batch_select
   - [x] execute
   - [x] exists
   - [x] index
   - [x] info
       - [x] policy passing `null` saw segfault. Excluded it for now.
   - [x] llist
   - [x] log
   - [x] operate
       - [x] `'should prepend and read a bin'` failing
   - [x] query
   - [x] remove
   - [x] scan
   - [x] udf_register
       - [ ] test with additional `language.LUA` fails
   - [x] udf_remove

**Wrap-up**
- [x] cleanup according to [comment](https://github.com/aerospike/aerospike-client-nodejs/pull/106/files#r51093917)
- [x] not have hardcoded OK checks according to [comment](https://github.com/eljefedelrodeodeljefe/aerospike-client-nodejs/commit/39decdc13c003fed5aa55cb72fac2a15af585709#commitcomment-15726973)

**Pending ToDos**
* [ ] no test for geo json and 2d index -> could not check accordingly
* [ ] llist test does neither skip nor can I make it run due to `before`  info condiation check. Info implementation wrong?
* [ ] some smaller tests are not working, see above.
* [ ] `lib/llist` could be refactored


lib/aerospike.js
```diff
@@ -235,62 +234,413 @@ var prepend = function prepend (key, bins, metadata, policy, callback) {
...
   if (client === null) {
     throw new Error('Client object creation failed - null value returned')
   }
 
-  var proto = Object.getPrototypeOf(client)
-  if (!proto.createQuery) {
-    proto.createQuery = proto.query
-    proto.query = query
-  }
+  this.proto = Object.getPrototypeOf(client)
```
jhecking: What is this needed for?

eljefedelrodeodeljefe: This is just wrapped what was before. I am sure this will disappear, but I haven't touched it yet. Further down you'll see `return client.`. This would be the mechanism to wrap the client in a `this._currentClient` in order to do two things:
* wrapping client to handle errors in class
* combining client creation and `client.connect()`

test/put.js
```diff
@@ -174,51 +175,49 @@ describe('client.put()', function () {
...
+  //       expect(record1.val).to.equal(record.val)
+  //       expect(record1.dval).to.equal(record.dval.Double)
+  //
+  //       Aerospike.remove(key, function (err, key) {
+  //         if (err) { throw new Error(err.message) }
+  //         done()
+  //       })
+  //     })
+  //   })
+  // })
```
eljefedelrodeodeljefe: @jhecking can you explain why this tests fails? Haven't touched any specifics here, but when I used it even with the 1.X.X Version Doubles were failing really badly. Leaving me to forcefully restart the vagrant machine. Since this would be quite the bad error in `prod` I am really curious.

jhecking: What server version are you testing against? Support for double data type was introduced in server version 3.6. Older servers don't handle unsupported data types well and will crash if they receive double values.

You can send an info request and check that the `features` field contains an entry `float`.

test/generators/put.js
```diff
@@ -23,10 +23,10 @@ function put_done (total, done) {
 
   return function (_key, record, metadata) {
     return function (err, key, skippy) {
-      switch (err.code) {
-        case status.AEROSPIKE_OK:
+      switch (err) {
```
jhecking: Can replace `switch` statement with an `if .. else` statement since we only check for the presence of the error.

test/generators/put.js
```diff
@@ -53,7 +53,7 @@ function put (client, n, keygen, recgen, metagen, done) {
     var metadata = metagen(key)
     var record = recgen(key, metadata)
     var callback = d(key, record, metadata)
-    client.put(key, record, metadata, callback)
+    Aerospike.put(key, record, metadata, callback)
```
jhecking: The generator doesn't need an instance of the client class in this case any more. So we can remove that from the method signature.

eljefedelrodeodeljefe:
@jhecking Okay, we could slowly prepare merging if you like. I have some problems with some tests (see description), which I really cannot make run. Can you advise? Might be my vagrant setup (?).

jhecking:
Sounds good. I will take a look. However I am travelling from tomorrow until next Wednesday so I probably won't be able to start merging until then.

If you could rebase on master that would be great. I did some refactoring on the tests so that all the test run using a single client instance. So you can remove all the `Aerospike.connect()` calls from the tests. There is a global `before` hook in the new `test_helper.js` file which sets up the client.

Cheers,
Jan

eljefedelrodeodeljefe:
Alright. Looking forward.

eljefedelrodeodeljefe:
@jhecking this should rebased now properly. The branch holds your commits now also though. Tell me if you need any other way.

I alos added 822ccd8 which would preserve the call to .client() in js by returning the class instance. This could be used in order to use the API as it is now. Personally I wouldn't use this though. 

Some work left to do, but we are getting there.

jhecking:
@eljefedelrodeodeljefe, this looks pretty good already. I have created a new [`api-refactor`](https://github.com/aerospike/aerospike-client-nodejs/tree/api-refactor) branch from master and merged your changes in. I also got all the unit tests working again.

I'm going to spend some more time testing this branch next week. There are also some more minor changes I want to make to the new API before I'm ready to merge this back to master.

eljefedelrodeodeljefe:
@jhecking awesome. tell me when I can be of help.

PR #111, opened 2016-02-10
Method Name Change

srinivasiyer:
Just fixed a small typo where remove was typed in a delete

### DIFF
--- a/docs/largelist.md
+++ b/docs/largelist.md
@@ -207,7 +207,7 @@ The parameters for `callback` argument:
 
 Example:
 ```js
-llist.delete({'key': 'ldt_key'}, function (error, response){
+llist.remove({'key': 'ldt_key'}, function (error, response){
   if (error && error.code !== status.AEROSPIKE_OK) {
     // handle failure
   } else {


jhecking:
Thanks for the fix, @srinivasiyer!

Please note that the new [List Data Type](http://www.aerospike.com/docs/guide/cdt-list.html), with it's rich set of operations, might be a better solution for your needs than the older `LargeList` functionality. Depends on how large your lists are going to get. Please let me know if you have questions about the difference between the two data types. Or feel free to post to our community forum: https://discuss.aerospike.com/.

srinivasiyer:
Hi Jan,

I did check the list datatypes as well. They are quite useful to manipulate
upon and the API is quite simple. The reason I chose Large List was because
the space allotted to them is not part of the record quota. Meaning if
there is a write-block-size of 1MB(which is the max limit), it applies that
rule to the aggregate of all bins in the record.

However, the large lists are excluded from the size quota of the record. In
certain cases I directly store binary data into records. While they usually
are not very large in size, they give me the safety buffer of not exceeding
the record size in case there are multiple binaries within the record which
total up to more than 1MB.

Please correct me here if my assumption is inaccurate.

Srinivas

jhecking:
Hi Srinivas,

Yes, the `write-block-size` limit is the main reason to use LList over List data type if the list might get too large to fit into a single record.

Just wanted to make sure you are aware of all the options!

Cheers,
Jan

PR #112, opened 2016-02-11
Test latest stable + show travis badge

revington:
`stable` alias has been deprecated. 

From nvm documentation:
>  - `stable`: this alias is deprecated, and only truly applies to `node` `v0.12` and earlier. Currently, this is an alias for `node`.


jhecking:
Thanks, @revington!

PR #122, opened 2016-04-27
Add support for node v6 by bumping nan dependency

djMax:
I'll leave it to you all to bump version in package.json and reflect as desired in History.md


djMax:
Not sure what to make of that failure on node v4.

jhecking:
This has been merged except for 9a9d15c since the `node` target already covers v6.

Issue #126, opened 2016-05-04
Query & Scan operations should return the record key


jhecking:
`Query#foreach` and `Scan#foreach` should return the record key. Currently the `data` event of `RecordStream` only includes the record's bins and meta data. A `Key` instance should be passed as the 3rd parameter:

    var stream = client.query('test', 'demo').foreach()
    stream.on('data', (bins, meta, key) => {
      // ...
    })

Note that by default the server only stores the key digest unless `Aerospike.policy.key.SEND` is used, so the `key` property of the returned Key object might be undefined and only `digest` will be populated.

PR #127, opened 2016-05-04
support return key in query

sel-fish:
v2.0.3 not support return key info in query.


jhecking:
This is already fixed in #126 and will be included in the next patch release v2.0.4.

sel-fish:
I found this issue [here](http://stackoverflow.com/questions/36998099/how-to-fetch-primary-key-in-aerospike-node-js-client/37017675#comment61597042_37017675), just read your response after I pull this merge request. 
Of course it is duplicate.

jhecking:
@sel-fish, sorry for the wasted effort. I had started on the fix right after I replied to your comment on Stack Overflow. Your pull request is still appreciated!

sel-fish:
@jhecking , it's so nice of you to say. Thanks for the fix :)

PR #149, opened 2016-10-10
Added Ubuntu 16 support

behcet:


behcet:
> /home/travis/build/aerospike/aerospike-client-nodejs/aerospike-server/instance1/share/libexec/aerospike-start: line 70: ulimit: open files: cannot modify limit: Operation not permitted

while the config has the following line
>  "sudo": false,



Changes works on my machine™.

jhecking:
Ubuntu 16.04 is supported, and has been since May, with the current v2 client version. Is there a specific reason that you are still using a v1 client?

behcet:
I have a legacy app that still is on 0.10

jhecking:
I've merged the PR and fixed the issue with the Travis CI setup (an incompatibility with the latest server release 3.10). But at this point I am not planning on releasing any further v1.x package updates to npmjs.com.

behcet:
Thanks for the merge but withholding releases is not going help me solve my problem. I understand this is an outdated version of the client but some long term support should be considered.

jhecking:
I have figured out how to use npm tags to avoid having this release show up as the latest release on npmjs.com and have released [aerospike@1.0.58](https://github.com/aerospike/aerospike-client-nodejs/releases/1.0.58).

behcet:
Thanks!

PR #177, opened 2017-04-05
Upgrade to Standard v10

jhecking:
Ref. https://github.com/feross/standard/blob/master/CHANGELOG.md#1000---2017-04-04


jhecking:
The new Buffer constructor APIs `Buffer.alloc()` and `Buffer.from()` where first implemented in Node.js v6 and later back-ported to Node.js v4.5 and v5.10. Switching to these new APIs means we will have to stop supporting older Node.js releases, in particular v0.12 and io.js.

jhecking:
With upcoming v2.5 release of the client I will announce the intent to drop support for Node.js v0.12 with the next client release (likely v3.0). For now I will revert the changes to the Buffer API and stick to Standard v9.

PR #179, opened 2017-04-05
Drop Node.js v0.12 support

jhecking:
Node.js v0.12 has officially reached end-of-life on 2016-12-31. (Ref. https://github.com/nodejs/LTS#lts-schedule) The Aerospike Node.js client will drop support for Node.js v0.12 as well as io.js with it's next major release v3.0. There is currently no concrete target release date for v3.0.


NawarA:
👍

JoshFerge:
LGTM

PR #201, opened 2017-05-31
Upgrade dependencies

benmccann:


benmccann:
The build is failing. It looks like https://github.com/aerospike/aerospike-client-nodejs/pull/179 is required

jhecking:
What problem are you trying to solve specifically?

Please note that a package dependency like `"mocha": "^3.0"` already allows any 3.x version of the package to be installed. Changing that to `"mocha": "^3.4"` is only necessary if there is a dependency on some feature that was introduced in `mocha@3.4` that is not available in versions 3.0-3.3 of the library.

We should probably upgrade the `yargs` dependency since it's really quite old. I'm not sure why the version number is pinned down like that for this package. But are you sure that upgrading from v1.x to v8.x does not break the benchmark or examples?

Last not least, I agree that we should allow minor version upgrades for `nan`, i.e. changing from a tilde range (e.g. `~2.3.0`) to a caret range (e.g. `^2.3`) is the right thing to do.

If you can keep the changes compatible to v0.12 for now, until #179 lands, I'd be happy to accept this PR. I know that `standard@10.0` requires Node.js v4 or later; not sure about the other packages you are trying to upgrade.

PR #274, opened 2018-09-03
Update standard package to v12

jhecking:
Update from v11 to v12 and run standard --fix to correct some minor rule
violations.  Ref. [standard changelog](https://github.com/standard/standard/blob/master/CHANGELOG.md#1200---2018-08-28)

Note: standard v12 drops support for Node.js 4 and only supports Node.js 6 and later.


jhecking:
Rebased on master.

Issue #286, opened 2018-10-30
Can't get to work on Alpine


aknuds1:
I'm trying to install aerospike-client-nodejs on Alpine Linux, using C library built from source, but it fails due to a mysterious relocation error. Please help me figure this one out!

The below Dockerfile reproduces the error (given a package.json file in the current directory):

```
FROM node:10-alpine
MAINTAINER Me

WORKDIR /app

ENV PREFIX=/opt/aerospike-client-c

RUN apk add --no-cache python2 make gcc g++ git libtool autoconf automake linux-headers \
openssl-dev zlib-dev libuv-dev

RUN git clone https://github.com/aerospike/aerospike-client-c.git && cd aerospike-client-c && \
git submodule update --init && \
make EVENT_LIB=libuv && \
mv target/* /opt/aerospike-client-c

COPY package.json .

RUN npm install aerospike
RUN node -e "require('aerospike')"
```

You should see an error like

```
/app/node_modules/bindings/bindings.js:88
        throw e
        ^

Error: Error relocating /app/node_modules/aerospike/build/Release/aerospike.node: sk_free: symbol not found
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:717:18)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
    at Module.require (internal/modules/cjs/loader.js:636:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at bindings (/app/node_modules/bindings/bindings.js:81:44)
    at Object.<anonymous> (/app/node_modules/aerospike/lib/aerospike.js:19:31)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
```

jhecking:
The client used to build/run on Alpine Linux in the past. However, I can reproduce the issue using your Dockerfile. There are at least two issues that are currently causing the build to fail:

(1) In `binding.gyp` I need to add `-lssl` to link against libssl explicitly.
(2) The Aerospike C client recently added a call to `pthread_attr_setaffinity_np` - this function does not seem to be supported on Alpine Linux.

I'll fix (1) but I'll need to talk to our C client developer re (2).

I've temporarily patched both issues in my docker image and am able to run the client.

[...]

PR #288, opened 2018-10-31
Fix Linux build by linking to ssl

aknuds1:
Link to ssl on Linux, to resolve linking problem that otherwise arises on Alpine Linux. See #286.


jhecking:
To be clear, this specifically affects Alpine Linux, not Linux in general.

aknuds1:
@jhecking I was wondering about this. Why isn't it necessary to link against libssl on other Linuxes?

PR #301, opened 2019-08-01
Upgrade to C client v4.6.4

jhecking:


jhecking:
Skipping C client v4.6.4 due to an issue with async info commands (CLIENT-1117) and moving straight to v4.6.5 instead.

PR #343, opened 2020-01-08
Update CHANGELOG.md

mcoberly2:


jhecking:
Isn't it kind of obvious that "server" is referring to the Aerospike server?

PR #393, opened 2020-12-18
fix: tls and tlsname mismatch

fschaeffler:
As stated in this TLS PR here #382, the attribute is called `tlsname`. The current version of the `config.cc` also makes use of `tlsname` like e.g. here in [src/main/config.cc](https://github.com/aerospike/aerospike-client-nodejs/blob/master/src/main/config.cc#L77).

However, in `lib/utils.js` [here](https://github.com/aerospike/aerospike-client-nodejs/blob/719b259c81a613a13424d4f133b5d9a8adc53d41/lib/utils.js#L37), the attribute is `tls` and not `tlsname`.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -34,7 +34,7 @@ function parseHostString (hostString) {
 
   const tlsName = parts[2]
   if (tlsName) {
-    host.tls = tlsName
+    host.tlsname = tlsName
   }
 
   const port = parts[3]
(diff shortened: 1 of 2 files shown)


jhecking:
Thanks, good catch! However, the `parseHostString` where the incorrect `tls` property name is used, is only called from the `Client#info` method, which is deprecated and likely going to be removed soon. So in practice this issue is unlikely to affect anyone.

fschaeffler:
It kinda affected us because I was searching in the source code on how to implement TLS. Mainly because the docu didn't give me any hints. I then stumbled upon the definition of `tls` in `parseHostString` and made use it. :see_no_evil:

jhecking:
You are right, the API docs could need some improvement to explain better how to connect to a cluster using TLS. I've filed https://github.com/aerospike/aerospike-client-nodejs/issues/394 as a reminder to improve this.

PR #421, opened 2021-06-26
Fix checking build result on Windows

bit0r1n:
Using the old method, you can build a package only with english MSBuild localization. This commit fix it.


jhecking:
Thanks!

PR #444, opened 2022-02-02
Aerospike Expressions Support & Build Changes for 4.0

rpandian-spike:


lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+
+/*********************************************************************************
+ * VALUE EXPRESSIONS
+ *********************************************************************************/
+
+const _valueExp = (op, valName) => (val) => [{ op, [valName]: val }]
+
+const _valueExpBytes = (op, valName, sizeName) => (val, size) => [{ op, [valName]: val, [sizeName]: size }]
+
+exports.bool = _valueExp(exp.ops.VAL_BOOL, 'boolVal')
```
jhecking, line 36: Missing documentation for all exported values.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
+// *****************************************************************************
+// Copyright 2021 Aerospike, Inc.
```
jhecking, line 2: Need to update Copyright statement to include 2022 in all changed files.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+
+/*********************************************************************************
+ * ARITHMETIC EXPRESSIONS
+ *********************************************************************************/
+
+const _VAargs = (op) => (...expr) => [].concat(
+  { op },
+  ...expr,
+  { op: exp.ops.END_OF_VA_ARGS }
+)
```
jhecking, line 190: Duplicate of `_VAExp` defined above for the logical expressions.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+exports.add = _VAargs(exp.ops.ADD)
+exports.sub = _VAargs(exp.ops.SUB)
+exports.mul = _VAargs(exp.ops.MUL)
+exports.div = _VAargs(exp.ops.DIV)
+
+const _p3Exp = (op) => (p1, p2) => [
+  { op: op, count: 3 },
+  ...p1,
+  ...p2
+]
```
jhecking, line 201: Can replace this as well as `_p2Exp` with something generic like this:

```javascript
const _paramsExp = (op) => (...params) => [
  { op, count: params.length + 1 },
  ...[].concat(...params)
]
```

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+ *********************************************************************************/
+exports.cond = _VAargs(exp.ops.COND)
+exports.let = _VAargs(exp.ops.LET)
+
+/*********************************************************************************
+ * LIST READ EXPRESSIONS
+ *********************************************************************************/
+
+function getContextType (ctx, type) {
+  return type
```
jhecking, line 245: Need to replace this with the same logic as the C client's [as_exp_get_ctx_type function](https://github.com/aerospike/aerospike-client-c/blob/master/src/main/aerospike/as_exp.c#L425-L435).

Also need to add unit test covering different list contexts, so that this function gets covered.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+}
+
+function getListType (type, returnType, isMulti) {
+  let expected = type
```
jhecking, line 249: Need to add some more unit tests for different list types/return types to ensure this function gets covered.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+  ..._int(op)
+]
+
+const _listStart = (ctx, op, param) => [
+  { op: exp.ops.CALL_VOP_START, count: 1 + param, ctx },
+  ..._int(op)
+]
+
+exports.lists = {
+  size: (bin, ctx = null) => [
```
jhecking, line 300: Need to add more unit tests to cover all the list expressions.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+    ...bin
+  ]
+}
+
+/*********************************************************************************
+ * MAP MODIFY EXPRESSIONS
+ *********************************************************************************/
+
+function getMapType (type, returnType, isMulti) {
+  let expected = type
```
jhecking, line 550: Need to add more unit tests using different map types/return types, to ensure this function gets covered.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+  { op: exp.ops.CALL, count: 5 },
+  ..._rtype(getContextType(ctx, exp.type.MAP)),
+  ..._int(exp.sys.CALL_CDT | exp.sys.FLAG_MODIFY_LOCAL),
+  { op: exp.ops.CALL_VOP_START, count: 1 + param + (policy ? extraParam : 0), ctx },
+  ..._int(op)
+]
+
+exports.maps = {
+
+  put: (bin, val, key, policy = null, ctx = null) => [
```
jhecking, line 604: Need to add unit tests covering all the map expressions.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+
+const _bitReadStart = (returnType, op, param) => [
+  ..._bitRead(returnType),
+  { op: exp.ops.CALL_VOP_START, count: 1 + param },
+  ..._int(op)
+]
+
+exports.bit = {
+
+  reSize: (bin, flags, byteSize, policy = null) => [
```
jhecking, line 933: Need to add unit tests to cover all bitwise expressions.

lib/exp.js
```diff
@@ -0,0 +1,1206 @@
...
+
+const _hllReadStart = (returnType, op, param) => [
+  ..._hllRead(returnType),
+  { op: exp.ops.CALL_VOP_START, count: 1 + param },
+  ..._int(op)
+]
+
+exports.hll = {
+
+  initMH: (bin, mhBitCount, indexBitCount, policy = null) => [
```
jhecking, line 1121: Need to add unit tests to cover all HLL expressions.

lib/exp_operations.js
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * const Aerospike = require('aerospike')
+ * const op = Aerospike.operations
+ * const key = new Aerospike.Key('test', 'demo', 'mykey1')
+ *
+ * var ops = [
+ *   op.append('a', 'xyz'),
+ *   op.incr('b', 10),
+ *   op.read('b')
+ * ]
```
jhecking, line 37: Update this example to use read/write operations using Aerospike Expressions.

lib/exp_operations.js
```diff
@@ -0,0 +1,103 @@
...
+// limitations under the License.
+// *****************************************************************************
+
+'use strict'
+
+/**
+ * @module aerospike/exp_operations
+ *
+ * @description This module provides functions to easily define operations to
+ * be performed on a record via the {@link Client#operate} command.
```
jhecking, line 23:
```suggestion
 * @description This module provides functions to easily create read/write operations using
 * Aerospike Expressions to be performed on a record via the {@link Client#operate} command.
```

lib/exp_operations.js
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @see {@link Client#operate}
```
jhecking, line 25:
```suggestion
 * @see {@link Client#operate}
 * @see {@link Aerospike#exp}
```

lib/exp_operations.js
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @classdesc Base class for all operations executed with the {@link
+ * Client#operate} command.
+ *
+ * Operations can be created using the methods in one of the following modules:
+ * * {@link module:aerospike/exp_operations} - General operations on all types.
+ * * {@link module:aerospike/operations} - General operations on all types.
+ * * {@link module:aerospike/lists} - Operations on CDT List values.
+ * * {@link module:aerospike/maps} - Operations on CDT Map values.
+ * * {@link module:aerospike/bitwise} - Operations on Bytes values.
```
jhecking, line 66: Please update. This description is just copied from the base Operation class and doesn't make sense here.

lib/exp_operations.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+class ExpOperation {
+  /**
+   * @protected
+   */
+  constructor (op, bin, exp, flags, props) {
```
jhecking, line 72: `props` argument is never used and can be removed. Alternatively, you can skip the whole `ExpOperation` class and just use the base `Operation` class instead, and pass `exp` and `flags` as extra properties, i.e.

```javascript
exports.read = function (bin, exp, flags) {
   return new Operation(ops.READ, bin, { exp, flags })
}
```

Using a separate class to extend `Operation` really only makes sense if you need to add additional functionality like the `withContext` functions for the List and Map operations. That's not the case for the Expression operations.

lib/policies/query_policy.js
```diff
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright 2013-2020 Aerospike, Inc.
+// Copyright 2013-2021 Aerospike, Inc.
```
jhecking, line 2: Can be updated to 2022 now. :-)

package.json
```diff
@@ -29,38 +29,41 @@
...
     "lint": "standard",
     "posttest": "npm run lint",
-    "coverage": "nyc npm test -- --reporter dot && nyc --reporter=lcov report",
+    "coverage": "nyc npm test && nyc --reporter=lcov report",
     "codecov": "codecov",
     "prepush": "npm test",
     "apidocs": "jsdoc -c jsdoc.json",
     "cppcheck": "cppcheck --quiet --enable=warning,style -I src/include src/main/",
-    "valgrind": "valgrind node ./node_modules/mocha/bin/_mocha -R dot -g '#slow' -i",
-    "prepare": "husky install"
```
jhecking, line 42: Why remove the Husky Git hooks?

package.json
```diff
@@ -29,38 +29,41 @@
...
     "bindings": "^1.3.0",
+    "change-case-all": "^1.0.14",
     "minimatch": "^3.0.4",
     "nan": "^2.14.0"
   },
   "devDependencies": {
     "chai": "^4.2.0",
     "choma": "^1.2.1",
-    "codecov": "^3.8.1",
+    "codecov": "^3.7.1",
```
jhecking, line 52: Any reason you are downgrading to an earlier version?

package.json
```diff
@@ -29,38 +29,41 @@
...
     "prepush": "npm test",
     "apidocs": "jsdoc -c jsdoc.json",
     "cppcheck": "cppcheck --quiet --enable=warning,style -I src/include src/main/",
-    "valgrind": "valgrind node ./node_modules/mocha/bin/_mocha -R dot -g '#slow' -i",
-    "prepare": "husky install"
+    "valgrind": "valgrind node ./node_modules/mocha/bin/_mocha -R dot -g '#slow' -i"
   },
   "dependencies": {
     "bindings": "^1.3.0",
+    "change-case-all": "^1.0.14",
```
jhecking, line 45: Where are you using this module?

package.json
```diff
@@ -29,38 +29,41 @@
...
     "deep-eql": "^4.0.0",
     "dirty-chai": "^2.0.1",
-    "husky": "^6.0.0",
-    "ink-docstrap": "^1.3.0",
-    "jsdoc": "^3.6.7",
-    "mocha": "^7.1.2",
+    "ink-docstrap": "^1.3.2",
+    "jsdoc": "^3.6.10",
+    "markdown-it": "^12.3.2",
+    "marked": "^4.0.10",
```
jhecking, line 58: Why do we need two separate Markdown parsers? And how are they being used?

package.json
```diff
@@ -29,38 +29,41 @@
...
+    "mocha": "^9.2.0",
     "mocha-clean": "^1.0.0",
     "nyc": "^15.0.0",
     "p-throttle": "^3.1.0",
+    "postcss": "^8.4.5",
+    "sanitize-html": "^2.6.1",
```
jhecking, line 64: How are these packages being used?

src/main/aerospike.cc
```diff
@@ -93,10 +94,10 @@ NAN_METHOD(setDefaultLogging)
 		if (log_from_jsobject(&g_log_info, info[0].As<Object>()) == AS_NODE_PARAM_OK) {
 			if (g_log_info.level < 0) {
 				// common logging does not support log level "OFF"
-				as_log_set_level(AS_LOG_LEVEL_ERROR);
-				as_log_set_callback(NULL);
+				as_log_set_level(AS_LOG_LEVEL_TRACE);
+				as_log_set_callback(as_log_callback_fnct);
 			} else {
-				as_log_set_level(g_log_info.level);
+				as_log_set_level(AS_LOG_LEVEL_TRACE);
```
jhecking, line 100: Were these changes you did just for debugging? Need to revert them?

src/main/policy.cc
```diff
@@ -59,9 +60,6 @@ int infopolicy_from_jsobject(as_policy_info* policy, Local<Object> obj, const Lo
 	if ((rc = get_optional_uint32_property(&policy->timeout, NULL, obj, "timeout", log)) != AS_NODE_PARAM_OK) {
 		return rc;
 	}
-	if ((rc = get_optional_uint32_property(&policy->timeout, NULL, obj, "totalTimeout", log)) != AS_NODE_PARAM_OK) {
-		return rc;
-	}
```
jhecking, line 64: The `timeout` property is used here just as a fallback to remain backwards compatibility. As mentioned earlier, I think we should stick to using the same name `totalTimeout` as the other (C/Java) clients. We should removed the fallback to also check the `timeout` property from the 4.0 release.

rpandian-spike: infoPolicy has timeout and basePolicy has total_timeout fields, so you mean to say use corresponding names "timeout" and "totalTimeout" from user objects?

jhecking: Yes. The base policy has two timeout values, i.e. `totalTimeout` and `socketTimeout`. Just using `timeout` might be a bit ambiguous.

lib/exp.js
```diff
@@ -29,29 +29,76 @@ const BIN_TYPE_UNDEF = 0
...
-const _valueExp = (op, valName) => (val) => [{ op, [valName]: val }]
+const _valueExp = (op, valName) => (value) => [{ op, [valName]: value }]
 
-const _valueExpBytes = (op, valName, sizeName) => (val, size) => [{ op, [valName]: val, [sizeName]: size }]
+const _valueExpBytes = (op, valName, sizeName) => (value, size) => [{ op, [valName]: value, [sizeName]: size }]
 
+/**
+ * Create boolean value.
+ *
+ * @param value boolean value.
```
jhecking, line 39: Would be best to specify the parameter type as well, using e.g.

```
@param {boolean} value - boolean value.
```

Ref. https://jsdoc.app/tags-param.html

lib/exp.js
```diff
@@ -64,40 +111,143 @@ const _keyTypeExp = (type) => () => [
...
+ * @return (integer value) Integer value of the key if the key is an integer.
+ * @ingroup expression
+ */
 exports.keyInt = _keyTypeExp(exp.type.INT)
 
+/**
+ * Create expression that returns the key as an string. Returns 'unknown' if
+ * the key is not a string.
+ *
+ * @return (string value) String value of the key if the key is a string.
```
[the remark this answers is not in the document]
jhecking: Also, the return type is not really `string` in this case, but an Aerospike Expression! We don't have a type specified for that, though. We should define an opaque type just for the docs, that we can then reference, e.g.

```
@typedef {Object} exp~AerospikeExpression

// ..

@returns {AerospikeExpression} Aerospike Expression representing the string value of the key, if the key is a string.
```

(Maybe can pick a shorter name than `AerospikeExpressions`...)

lib/exp_operations.js
```diff
@@ -19,20 +19,27 @@
...
+ * @see {@link Client#exp}
  *
  * @example
  *
  * const Aerospike = require('aerospike')
  * const op = Aerospike.operations
+ * const exp = Aerospike.expressions
  * const key = new Aerospike.Key('test', 'demo', 'mykey1')
- *
+ * const tempBin = 'ExpVar'
```
jhecking, line 35: This constant `tempBin` is used exactly once in this example - I'm not sure that it helps with readability, as it's not clear what its purpose is.

lib/exp.js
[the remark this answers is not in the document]
rpandian-spike: will add these in stages

lib/exp.js
```diff
@@ -762,10 +762,6 @@ exports.var = (varName) => [
  * LIST READ EXPRESSIONS
  *********************************************************************************/
 
-function getContextType (ctx, type) {
```
jhecking: I think this function is required, and needs to be implemented to match the logic in the corresponding C function here: https://github.com/aerospike/aerospike-client-c/blob/master/src/main/aerospike/as_exp.c#L425-L435

jhecking: Also means you need unit tests to cover the different context types used in list expressions.

lib/exp.js
```diff
@@ -0,0 +1,2655 @@
...
+      expected = exp.type.INT
+      break
+    case lists.returnType.VALUE:
+      if (isMulti) {
+        expected = exp.type.LIST
+      }
+      break
+    case lists.returnType.NONE:
+    default:
+      console.log('getListType valueType:' + type + 'returnType:' + returnType + ' FAILED')
```
jhecking, line 786: Instead of logging and returning an `ERROR` type, maybe the client should just raise a `TypeError` here. What do you think?

rpandian-spike: sure will add that

package.json
```diff
@@ -39,13 +51,15 @@
...
+    "prepare": "husky install",
+    "__comment__": "prepare: husky install"
   },
   "dependencies": {
     "bindings": "^1.3.0",
     "minimatch": "^3.0.4",
     "nan": "^2.14.0",
-    "node-gyp": "^8.4.1"
+    "node-gyp": "^8.4.1",
+    "node-pre-gyp": "^0.17.0"
```
jhecking, line 62: Should use `@mapbox/node-pre-gyp` instead:

> On Feb 9th, 2021 @mapbox/node-pre-gyp@1.0.0 was [released](https://github.com/mapbox/node-pre-gyp/blob/HEAD/CHANGELOG.md). Older, unscoped versions that are not part of the @mapbox org are deprecated and only @mapbox/node-pre-gyp will see updates going forward.

https://www.npmjs.com/package/@mapbox/node-pre-gyp

lib/exp.js
```diff
@@ -783,11 +783,17 @@ function getListType (type, returnType, isMulti) {
       break
     case lists.returnType.NONE:
     default:
-      console.log('getListType valueType:' + type + 'returnType:' + returnType + ' FAILED')
+      console.log("getListType valueType:"+type+"returnType:"+returnType+" FAILED")
```
jhecking: Just a stylistic note: JavaScript template literals are the best way to embed expressions into strings now.

E.g. instead of

    "foo: " + foo + ", bar: " + bar

use

    `foo: ${foo}, bar: ${bar}`

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

package.json
```diff
@@ -28,39 +28,54 @@
...
-    "coverage": "nyc npm test -- --reporter dot && nyc --reporter=lcov report",
+    "coverage": "nyc npm test && nyc --reporter=lcov report",
     "codecov": "codecov",
     "prepush": "npm test",
     "apidocs": "jsdoc -c jsdoc.json",
     "cppcheck": "cppcheck --quiet --enable=warning,style -I src/include src/main/",
     "valgrind": "valgrind node ./node_modules/mocha/bin/_mocha -R dot -g '#slow' -i",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "__comment__": "prepare: husky install"
```
jhecking, line 55: Can remove this?

package.json
```diff
@@ -28,39 +28,54 @@
...
   },
   "devDependencies": {
     "chai": "^4.2.0",
     "choma": "^1.2.1",
-    "codecov": "^3.8.1",
+    "codecov": "^3.8.3",
     "deep-eql": "^4.0.0",
     "dirty-chai": "^2.0.1",
-    "husky": "^6.0.0",
-    "ink-docstrap": "^1.3.0",
```
jhecking, line 56: API docs won't build without this package:

```
❯ npm run apidocs

> aerospike@4.0.0 apidocs /Users/jhecking/aerospike/aerospike-client-nodejs
> jsdoc -c jsdoc.json

FATAL: Unable to load template: Cannot find module './node_modules/ink-docstrap/template/publish'
Require stack:
- /Users/jhecking/aerospike/aerospike-client-nodejs/node_modules/jsdoc/cli.js
- /Users/jhecking/aerospike/aerospike-client-nodejs/node_modules/jsdoc/jsdoc.js
```

package.json
```diff
@@ -28,39 +28,54 @@
...
+  "binary": {
+    "module_name": "aerospike",
+    "module_path": "./lib/binding/{configuration}/{node_abi}-{platform}-{arch}/",
+    "remote_path": "./{module_name}/v{version}/{configuration}/",
+    "package_name": "{module_name}-v{version}-{node_abi}-{platform}-{arch}.tar.gz",
+    "host": "https://www.npmjs.com/package/aerospike"
+  },
+  "config": {
+    "unsafe-perm": true
+  },
```
jhecking, line 40: Is this really required? unsafe-perm defaults to true for non-root users anyway: https://docs.npmjs.com/cli/v6/using-npm/config#unsafe-perm

package.json
```diff
@@ -28,39 +28,54 @@
...
   "scripts": {
-    "test": "mocha",
+    "preinstall": "npm install @mapbox/node-pre-gyp",
+    "install": "node-pre-gyp install --fallback-to-build",
+    "test": "mocha --exit",
```
jhecking, line 44: If you have to use `--exit` that strongly indicates that there's either a bug in the client or a problematic test which prevent the event loop from shutting down? We should find and fix those issues. Running mocha with `--exit` just hides the problem.

package.json
```diff
@@ -28,39 +28,54 @@
...
+    "codecov": "^3.8.3",
     "deep-eql": "^4.0.0",
     "dirty-chai": "^2.0.1",
-    "husky": "^6.0.0",
-    "ink-docstrap": "^1.3.0",
-    "jsdoc": "^3.6.7",
-    "mocha": "^7.1.2",
+    "husky": "^7.0.4",
+    "jsdoc": "^3.6.10",
+    "marked": "^4.0.10",
```
jhecking, line 72: What is this used for?

lib/client.js
```diff
@@ -625,7 +625,7 @@ Client.prototype.connect = function (callback) {
  *                   index: indexName,
  *                   type: indexType,
  *                   datatype: dataType }
- *   let policy = new Aerospike.InfoPolicy({ totalTimeout: 100 })
+ *   let policy = new Aerospike.InfoPolicy({ timeout: 100 })
```
jhecking, line 628: Should revert back to `totalTimeout` in all examples.

lib/policies/base_policy.js
```diff
@@ -35,14 +35,14 @@ class BasePolicy {
      *
      * If <code>socketTimeout</code> is not zero and the socket has been idle
      * for at least <code>socketTimeout</code>, both <code>maxRetries</code>
-     * and <code>totalTimeout</code> are checked. If <code>maxRetries</code>
-     * and <code>totalTimeout</code> are not exceeded, the transaction is
+     * and <code>timeout</code> are checked. If <code>maxRetries</code>
```
jhecking, line 38: Revert back to `totalTimeout` in the docs.

lib/policies/base_policy.js
```diff
@@ -53,22 +53,22 @@ class BasePolicy {
...
-     * If <code>totalTimeout</code> is zero, there will be no total time limit.
+     * If <code>timeout</code> is zero, there will be no total time limit.
      *
      * @type number
      * @default 1000
      */
     this.totalTimeout = props.totalTimeout
     if (this.totalTimeout === undefined) {
-      this.totalTimeout = props.timeout
+      this.totalTimeout = props.totalTimeout
```
jhecking, line 71: I think we can remove the fallback to use `timeout` if `totalTimeout` is not set. But as-is, this line is non-sensical since it's exactly the same as line 69 above.

lib/policies/info_policy.js
```diff
@@ -38,9 +38,9 @@ class InfoPolicy {
      *
      * @type number
      */
-    this.totalTimeout = props.totalTimeout
-    if (this.totalTimeout === undefined) {
-      this.totalTimeout = props.timeout
+    this.timeout = props.timeout
+    if (this.timeout === undefined) {
+      this.timeout = props.timeout
```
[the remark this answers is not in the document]
jhecking: Don't forget to call out this backward incompatible change in the release notes and at https://developer.aerospike.com/client/nodejs/usage/incompatible as well!

lib/policy.js
```diff
@@ -90,7 +90,7 @@ const WritePolicy = require('./policies/write_policy')
  * // Set default policies for read and write commands
  * const defaults = {
  *   socketTimeout: 50,
- *   totalTimeout: 100
+ *   timeout: 100
```
jhecking, line 93: Revert back to `totalTimeout`.

src/main/aerospike.cc
```diff
@@ -93,8 +94,8 @@ NAN_METHOD(setDefaultLogging)
 		if (log_from_jsobject(&g_log_info, info[0].As<Object>()) == AS_NODE_PARAM_OK) {
 			if (g_log_info.level < 0) {
 				// common logging does not support log level "OFF"
-				as_log_set_level(AS_LOG_LEVEL_ERROR);
-				as_log_set_callback(NULL);
+				as_log_set_level(AS_LOG_LEVEL_TRACE);
+				as_log_set_callback(as_log_callback_fnct);
```
jhecking, line 98: I really think you should revert this change! Else setting the log level to OFF on the JS side, will enable full TRACE level logging in the C client.

src/main/util/log.cc
```diff
@@ -122,6 +122,8 @@ _as_v8_log_function(const LogInfo* log, as_log_level level, const char* func,
 	char msg[1024];
 	vsnprintf(msg, 1024, fmt, args);
 
+	printf("%s\n", msg);
```
jhecking, line 125: Should remove this again?

test/apply.js
```diff
@@ -55,7 +55,7 @@ describe('client.apply()', function () {
 
   it('should invoke an UDF with apply policy', function (done) {
     const policy = new Aerospike.ApplyPolicy({
-      totalTimeout: 1500
+      timeout: 1500
```
jhecking, line 58: Revert to `totalTimeout`. In all other tests as well.

test/exp_map.js
```diff
@@ -0,0 +1,357 @@
...
+        //     exp.maps.getByValueRange(
+        //       exp.binMap('tags'),
+        //       exp.str('yellow'),
+        //       exp.str('blue'),
+        //       maps.returnType.COUNT),
+        //     0),
+        //   op.read('tags')
+        // ]
+        // const result = await client.operate(key, ops, {})
+        // console.log(result)
```
jhecking, line 117: Can remove all the commented out code from the tests?

test/util/options.js
```diff
@@ -49,7 +49,7 @@ const parser = yargs
     },
     log: {
       alias: 'l',
-      default: Aerospike.log.WARN,
+      default: Aerospike.log.TRACE,
```
jhecking, line 52: I think should revert the default to WARN?

jhecking (commented):
Looking good! 👏  Just few more minor changes required.

* package-lock.json file needs to be updated to be in sync with latest changes to package.json.
* Release notes need to be updated with new Aerospike Expressions features as well as build changes.
* jsdocs still have some issues, e.g. https://github.com/aerospike/aerospike-client-nodejs/pull/444/files#r802211630
* Ensure API docs can be build all links work as expected.
* Code coverage needs to be improved for all the new code added: https://app.codecov.io/gh/aerospike/aerospike-client-nodejs/compare/444

test/client.js
```diff
@@ -246,4 +246,21 @@ describe('Client', function () {
       })
     })
   })
+
+  context.only('minConnsPerNode', function () {
```
jhecking, line 250:
```suggestion
  context('minConnsPerNode', function () {
```

test/query.js
```diff
@@ -1,568 +0,0 @@
...
-const GEO2DSPHERE = Aerospike.indexDataType.GEO2DSPHERE
-const LIST = Aerospike.indexType.LIST
-const MAPVALUES = Aerospike.indexType.MAPVALUES
-const MAPKEYS = Aerospike.indexType.MAPKEYS
-
-const keygen = helper.keygen
-const metagen = helper.metagen
-const putgen = helper.putgen
-
-describe('Queries', function () {
```
jhecking, line 44: Any reason to remove all of these unit tests for queries, even though they don't use predexp or any other deprecated functionality?

src/main/expressions.cc
```diff
@@ -0,0 +1,150 @@
...
+	if (Nan::Has(entry_obj, Nan::New("listPolicy").ToLocalChecked()).FromJust()) {
+		printf("check listPolicy start");
+		Local<Value> policy_obj = Nan::Get(entry_obj, Nan::New("listPolicy").ToLocalChecked()).ToLocalChecked();
+		
+		if (policy_obj->IsObject() && 
+				get_optional_list_policy(entry->v.list_pol, NULL, policy_obj.As<Object>(), log)) {
+			printf("check listPolicy found");
+			return AS_NODE_PARAM_OK;
+		}
+		printf("check listPolicy failed");
```
dwelch-spike, line 116: Should these printf calls be in here?

src/main/expressions.cc
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (Nan::Has(entry_obj, Nan::New("mapPolicy").ToLocalChecked()).FromJust()) {
+		// TODO: implement policy convertion
+		return AS_NODE_PARAM_OK;
```
dwelch-spike, line 121: Looks like map policies still need to be converted.

src/main/expressions.cc
```diff
@@ -0,0 +1,150 @@
...
+	for (int i = 0; i < length; i++) {
+		*entry = { };
+		Local<Object> entry_obj = Nan::Get(entries_ary, i).ToLocalChecked().As<Object>();
+		if ((rc = convert_entry(entry_obj, entry, log)) != AS_NODE_PARAM_OK) {
+			as_v8_error(log, "Error converting expression entry: %i", i);
+			goto done;
+		}
+		entry++;
+	}
+	// TODO: destroy *filter_exp after use
```
dwelch-spike, line 144: Still need to do this?

lib/exp.js
dwelch-spike: Is this still an issue? Looks like the types need curly braces?

lib/exp.js
```diff
@@ -0,0 +1,2649 @@
...
+ * @return (integer or float value)
+ */
+exports.max = _VAExp(exp.ops.MAX)
+
+/*********************************************************************************
+ * FLOW CONTROL AND VARIABLE EXPRESSIONS
+ *********************************************************************************/
+/**
+ * Conditionally select an expression from a variable number of expression pairs
+ * followed by default expression action. Requires server version 5.6.0+.
```
dwelch-spike, line 721: Seems like the Control and variable expressions need tests?

dwelch-spike (commented):
@cwm-spike asked me to take a look at the expressions implementation here to see if it supports dynamic typing. What I mean is the ability to leave out expressions like exp.Int(), exp.Str etc in favor of their native type and have the expressions system work it out. For example `xp.eq(exp.binInt('intVal'), exp.int(42))` would become `xp.eq(exp.binInt('intVal'), 42)`. It looks like this isn't supported at the moment. Might be worth considering as a convenience feature for users.

Other comments are just a few things I noticed. Note I didn't do a deep dive, only skimmed expressions.

src/main/expressions.cc
jhecking: Should be converted into `as_v8_error` statements.

jhecking:
> @cwm-spike asked me to take a look at the expressions implementation here to see if it supports dynamic typing. What I mean is the ability to leave out expressions like exp.Int(), exp.Str etc in favor of their native type and have the expressions system work it out. For example `xp.eq(exp.binInt('intVal'), exp.int(42))` would become `xp.eq(exp.binInt('intVal'), 42)`. It looks like this isn't supported at the moment. Might be worth considering as a convenience feature for users.

Interesting suggestion! I think it would not be difficult to implement that. Just need to update the `compile_expression` and/or `convert_entry` functions in [src/main/expressions.cc](https://github.com/aerospike/aerospike-client-nodejs/blob/rp-dev/src/main/expressions.cc) to detect JS native null/Number/String/Boolean/Buffer objects and turn them into the right C data structures. Numbers are a bit tricky because JS doesn't distinguish between ints and floats, so if the user passes in `3.0` there's actually no way to tell whether it's supposed to be an int or float. But we have already dealt with that problem in other parts of the client by using a dedicated `Double` class to allow the user to make that distinction explicit.

I'm a bit worried whether this might lead to problems where the user e.g. passes in a bin name as a plain JS string rather than a `bin*` expression. But those kinds of errors are easily detectable. And overall the UX/DX benefits far outweigh the potential problems, I think.

PR #446, opened 2022-02-27
Add TypeScript typings

### LINKED ISSUE #311: Feature Request: TypeScript support
It'll be nice to have a TypeScript declaration alongside.

bit0r1n:
Resolves #311.


jhecking:
First off: Thank you! This looks great.

Out of curiosity, did you create the typings manually or is there some tool or script you are using?

Last not least, there are a lot of functions that are marked as `/** @private */` in the JSdocs. I think we should omit those functions from the Typescript typings as well.

jhecking:
@rpandian-spike Typescript bindings would make a great feature addition to the upcoming 4.0 release!

bit0r1n:
Yes, I did it manually, because I think that automatically generated typings will be inaccurate (like some methods, that described with JSDoc, returns only Promise, without typing result of promise, or some arguments marked as any, but actually it takes some specified JS and Aerospike types, and there are more such cases).
About privatizing, what to do with exporting objects with private methods/properties, like in InfoModule (likewise info.js) exporting "separators", but it marked as private.

jhecking:
> what to do with exporting objects with private methods/properties, like in InfoModule (likewise info.js) exporting "separators", but it marked as private.

Can decide on a case-by-case basis. E.g. in the case of `separators`, it was originally marked as private, and was only later exported. Should probably remove the `@private` declaration and include it in the TS typings.

bit0r1n:
There are some strange private-field usage:
- error.js - private static methods (fromASError, copyASErrorProperties)
- job.js - private static method (pollUntilDone)
- policy.js - exporting private method (createPolicy)
I left typings for these methods, because in case of 2 first files there is no private field for static methods, in case of 3rd file it exports method, that marked as private.

jhecking:
The use of the `@private` tag just indicates that these are considered internal functions, that should not be used clients of the Aerospike Node.js library. There is no guarantee of backwards compatibility and these functions could be changed or removed at any time. As such, I would also not include them in the TypeScript typings. Does that make sense?

jhecking:
One thing we could do, is to use the `@package` tag for these functions, rather than `@private`:

> The `@package` tag marks a symbol as package-private. Typically, this tag indicates that a symbol is available only to code in the same directory as the source file for this symbol. [[Source](https://jsdoc.app/tags-package.html)]

bit0r1n:
> One thing we could do, is to use the `@package` tag for these functions, rather than `@private`:
> 
> 
> 
> > The `@package` tag marks a symbol as package-private. Typically, this tag indicates that a symbol is available only to code in the same directory as the source file for this symbol. [[Source](https://jsdoc.app/tags-package.html)]

Yeah, it would be correct.

bit0r1n:
> The use of the `@private` tag just indicates that these are considered internal functions, that should not be used by clients of the Aerospike Node.js library. There is no guarantee of backwards compatibility and these functions could be changed or removed at any time. As such, I would also not include them in the TypeScript typings. Does that make sense?

In my understanding, JSDoc's `@private` marks member of class as private, so it can't be accessible outside of class methods.

jhecking:
JSDoc doesn't control accessibility, but is used just to generate documentation. Hence the meaning of the `@private` tag has a slightly different meaning compared to e.g. JavaScript's "new" [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields).

> The `@private` tag marks a symbol as private, or **not meant for general use**. Private members are not shown in the generated output[.]

[emphasis mine, [Source](https://jsdoc.app/tags-private.html)]

I see the Typescript typings as somewhere between the two, i.e. as a more precise definition of the _public_ API of the client. I don't think the typings need to include any private features, which a user of the client should not use anyway.

jhecking:
> > One thing we could do, is to use the `@package` tag for these functions, rather than `@private`:
> 
> Yeah, it would be correct.

Please feel free to change the `@private` tags to `@package` in the JSDocs where appropriate.

By default, symbols tagged as `@package` will be included in the generated documentation, but I believe this can be controlled using the `--access` [command line flag](https://jsdoc.app/about-commandline.html). So we could still limit the published API docs to only include `@public` symbols.

jhecking:
> So I can swap all `@private` tags to `@package`?

Honestly, I don't see that it makes any difference. But if you feel strongly about it, go ahead and swap it. If you do, please also add an `access` option to the jsdoc config to only include public and untagged symbols in the generated API docs:

```diff
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -28,6 +28,7 @@
     "destination": "./build/apidocs",
     "template": "./node_modules/ink-docstrap/template",
     "readme": "./docs/overview.md",
-    "tutorials" : "./docs/tutorials"
+    "tutorials": "./docs/tutorials",
+    "access": "public,undefined"
   }
 }
```

> I need to remove these methods/objects from typings?

That would be my preference. Unless there is some benefit to include private symbols in the typings that I'm maybe not aware of?

bit0r1n:
I think it’s worth leaving it as it is now, since I tried to cover everything that is declared, and I don’t really want to remove the typings for some methods due to the fact that the method is not declared in the documentation (tbh I would to open early mentioned private methods for documentation, but probably there is reason why they are hidden).

jhecking:
> I think it’s worth leaving it as it is now

Works for me. :-)

> tbh I would to open early mentioned private methods for documentation, but probably there is reason why they are hidden.

Some of these "private" methods actually do have some documentation, e.g. `Job.pollUntilDone`. But I agree that it would be good if some of the other private methods which are used across modules were documented as well, if only for the benefit of other people working on the client code.

As I mentioned, the main reason these methods are tagged `@private` is that they are not considered part of the public API of the client, meaning there is no guarantee of backward compatibility from one version to the next – even minor versions.

jhecking:
Merged! Thank you for your contribution to the Aerospike Node.js client, @bit0r1n. Much appreciated!

PR #454, opened 2022-03-28
PR: ClientNodeJS - Support partition scan

rpandian-spike:


lib/scan.js
```diff
@@ -136,6 +136,13 @@ function Scan (client, ns, set, options) {
    * @member {boolean} Scan#concurrent
    */
   this.concurrent = options.concurrent
+
+  /**
+   * If set to <code>true</code>, the scan will return only those belongs to partitions.
+   *
+   * @member {boolean} Scan#pfEnabled
+   */
+  this.pfEnabled = false
```
jhecking, line 145: Should this be `@private`? Else if some client sets `pfEnabled` to `true` directly, but does not set `partFilter`, that would not work, right?

src/main/commands/scan_async.cc
```diff
@@ -66,8 +69,19 @@ NAN_METHOD(AerospikeClient::ScanAsync)
 		as_v8_debug(log, "Using scan ID %lli for async scan.", scan_id);
 	}
 
-	as_v8_debug(log, "Sending async scan command");
-	status = aerospike_scan_async(client->as, &cmd->err, p_policy, &scan, &scan_id, async_scan_listener, cmd, NULL);
+	as_partition_filter_set_all(&pf);
+	if (partitions_from_jsobject(&pf, info[2].As<Object>(), log) == AS_NODE_PARAM_OK) {
```
jhecking, line 73: What if the partitions object is invalid? Shouldn't the client abort the whole operation in that case with an error, rather than just silently ignore it?

src/main/commands/scan_async.cc
```diff
@@ -66,8 +69,20 @@ NAN_METHOD(AerospikeClient::ScanAsync)
 		as_v8_debug(log, "Using scan ID %lli for async scan.", scan_id);
 	}
 
-	as_v8_debug(log, "Sending async scan command");
-	status = aerospike_scan_async(client->as, &cmd->err, p_policy, &scan, &scan_id, async_scan_listener, cmd, NULL);
+	as_partition_filter_set_all(&pf);
+	if (partitions_from_jsobject(&pf, &pf_defined, info[2].As<Object>(), log) != AS_NODE_PARAM_OK) {
+		CmdErrorCallback(cmd, AEROSPIKE_ERR_PARAM, "Policy object invalid");
```
jhecking, line 74:
```suggestion
		CmdErrorCallback(cmd, AEROSPIKE_ERR_PARAM, "Partitions object invalid");
```

src/main/policy.cc
```diff
@@ -317,3 +317,47 @@ int scanpolicy_from_jsobject(as_policy_scan* policy, Local<Object> obj, const Lo
...
+		if (!pf_obj->IsUndefined() && pf_obj->IsObject()) {
+			if ((rc = get_optional_uint32_property((uint32_t*) &pf->begin, NULL, pf_obj.As<Object>(), "begin", log)) != AS_NODE_PARAM_OK) {
+				return rc;
+			}
+			if ((rc = get_optional_uint32_property((uint32_t*) &pf->count, NULL, pf_obj.As<Object>(), "count", log)) != AS_NODE_PARAM_OK) {
+				return rc;
+			}
+			int len = AS_DIGEST_VALUE_SIZE;
+			bool digest_defined = false;
+			if ((rc = get_optional_bytes_property((uint8_t **)&pf->digest.value, &len, &digest_defined, pf_obj.As<Object>(), "digest", log)) != AS_NODE_PARAM_OK) {
```
jhecking, line 346: Should `digest` be a String or a Buffer? The docs for the `Scan#partitions` function specify the type of the `digest` parameter as `string`:

     @param {string} digest

jhecking (approved):
Some of the CI tests failed with a core dump:

> Fatal error: glibc detected an invalid stdio handle
> ...
> Aborted (core dumped)

https://github.com/aerospike/aerospike-client-nodejs/runs/5758202356?check_suite_focus=true

test/query.js
```diff
@@ -234,17 +234,22 @@ describe('Queries', function () {
...
+          let count = 0
+          stream.on('data', record => {
+            expect(++count).to.equal(1)
+            expect(record.key).to.be.instanceof(Key)
+            expect(record.key.key).to.equal(uniqueKey)
+          })
+          stream.on('end', done)
+        } catch (err) {
+          console.error('Partition query not supported by connected server!')
+          this.skip('Partition query not supported by connected server!')
```
jhecking, line 251: Please use `helper.skipUnlessVersion('>= X.Y.Z', this)` instead, e.g. https://github.com/aerospike/aerospike-client-nodejs/blob/5e50e517a5e4b1296162235cefbe44023a6031d3/test/query.js#L526

lib/exp.js
```diff
@@ -26,6 +26,45 @@ const cdtCtx = require('./cdt_context')
...
+ * @description module provides functions to
+ * create expressions for use in key operations.
+ *
+ * @see {@link Client#operate}
+ *
+ * @example
+ *
+ * const Aerospike = require('aerospike')
+ * const op = Aerospike.operations
+ * const exp = Aerospike.expressions
```
jhecking, line 42:
```suggestion
 * const exp = Aerospike.exp
```

lib/aerospike.js
```diff
@@ -48,8 +48,7 @@ exports.filter = require('./filter')
  * Client#operate} command.
  * @summary {@link module:aerospike/exp|aerospike/exp} module
  */
-const exp = exports.exp = require('./exp')
-exp.type = as.exp.type
```
jhecking, line 61: Won't this break this line further down?

    exp.operations = require('./exp_operations')

lib/policies/read_policy.js
```diff
@@ -77,6 +77,17 @@ class ReadPolicy extends BasePolicy {
      * @since v3.7.0
      */
     this.deserialize = props.deserialize
+
+    /**
+     * Should as_record instance be allocated on the heap before user listener is called in
+     * async commands. If true, the user is responsible for calling as_record_destroy() when done
+     * with the record. If false, as_record_destroy() is automatically called by the client after
+     * the user listener function completes. This field is ignored for sync commands.
```
jhecking, line 85: How is this supposed to work from JS-land? JS code can't call the `as_record_destroy` function unless you are going to expose some JS API for that?

rpandian-spike:
its done.

PR #455, opened 2022-03-30
PR: Client-nodeJS remove obsolete predexp feature

rpandian-spike:


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "aerospike-client-c"]
 	path = aerospike-client-c
 	url = https://github.com/aerospike/aerospike-client-c.git
-	branch = master
+	branch = stage
```
jhecking, line 4: Wouldn't it be best to point to a specific release tag instead? E.g. [`6.0.0`](https://github.com/aerospike/aerospike-client-c/releases/tag/6.0.0)?

lib/query.js
```diff
@@ -290,16 +263,12 @@ Query.prototype.select = function (bins) {
...
- * to limit the results returned by the query. This method takes either a
- * secondary index filter created using the {@link
- * module:aerospike/filter|filter module}, or a sequence of predicate
- * expressions created using the {@link module:aerospike/predexp|predexp
- * module} as argument. It can also be called multiple times to add both kinds
- * of filters to the same query. (Max. one of each kind.)
+ * @description Use a secondary index filter to limit the results returned by the query. 
+ * This method takes secondary index filter created using the {@link
+ * module:aerospike/filter|filter module} as argument. 
+ * It can also be called multiple times to add filters to the same query.
```
jhecking, line 269: This sentence should be removed. Only a single secondary filter predicate is supported.

lib/query.js
```diff
@@ -320,42 +289,28 @@ Query.prototype.select = function (bins) {
  * })
  *
  * @see {@link module:aerospike/filter} to create secondary index filters.
- * @see {@link module:aerospike/predexp} to create predicate expression filters.
  *
  * @example <caption>Applying a predicate expression in addition to a secondary
  * index filter to find only record where the 'tags' list bin contains the
  * value 'blue' but also contains at least one other value that is not
  * 'blue':</caption>
```
jhecking, line 296: Example maybe can be removed, since there already is a separate example showing how to apply secondary index filters.

lib/query.js
```diff
@@ -320,42 +289,28 @@ Query.prototype.select = function (bins) {
...
  *   stream.on('data', record => { console.info(record.bins.tags) })
  *   stream.on('error', error => { throw error })
  *   stream.on('end', () => client.close())
  * })
  */
 Query.prototype.where = function (predicate) {
-  if (Array.isArray(predicate) && predicate[0] instanceof predexp.PredicateExpression) {
-    this.setPredexpFilter(predicate)
-  } else if (predicate instanceof filter.SindexFilterPredicate) {
+  if (predicate instanceof filter.SindexFilterPredicate) {
```
jhecking, line 313: I think the client should raise a TypeError if the predicate is not a filter predicate.

jhecking (approved):
👍 Looks good!

.gitmodules
rpandian-spike: here you can specify only branch but with client-c commit hash we can point to particular release hash.

PR #461, opened 2022-04-19
PR: Client-nodejs BatchWrite Support

rpandian-spike:


lib/client.js
```diff
@@ -391,6 +394,254 @@ Client.prototype.batchRead = function (records, policy, callback) {
...
+ *   if (error) throw error
+ *   client.batchRemove(batchRecords, function (error, results) {
+ *     if (error) throw error
+ *     results.forEach(function (result) {
+ *       console.log(result)
+ *     })
+ *   })
+ * })
+ */
+Client.prototype.batchRemove = function (records, batchPolicy, batchRemovePolicy, callback) {
```
jhecking, line 629: Likewise, what can I do using the `batchRemove` that I can't also do using `batchWrite`? Why would I use one vs the other? I think the docs should answer that question.

lib/policies/base_policy.js
```diff
@@ -87,21 +87,24 @@ class BasePolicy {
     this.maxRetries = props.maxRetries
 
     /**
-     * Use zlib compression on write or batch read commands when the command
-     * buffer size is greater than 128 bytes. In addition, tell the server to
-     * compress it's response on read commands. The server response compression
-     * threshold is also 128 bytes.
+     * Milliseconds to sleep between retries.  Enter zero to skip sleep.
+     * This field is ignored when max_retries is zero.
```
jhecking, line 91:
```suggestion
     * This field is ignored when maxRetries is zero.
```

lib/policies/base_policy.js
```diff
@@ -87,21 +87,24 @@ class BasePolicy {
...
+     * This field is also ignored in async mode.
```
jhecking, line 92: Node.js client is always using async mode so does it even make sense to include this option?

lib/policies/base_policy.js
```diff
@@ -87,21 +87,24 @@ class BasePolicy {
...
+     * Reads do not have to sleep when a node goes down because the cluster
+     * does not shut out reads during cluster reformation.  The default for
+     * reads is zero.
      *
-     * Requires Enterprise Server version >= 4.8.
+     * The default for writes is also zero because writes are not retried by default.
+     * Writes need to wait for the cluster to reform when a node goes down.
+     * Immediate write retries on node failure have been shown to consistently
+     * result in errors.  If max_retries is greater than zero on a write, then
+     * sleep_between_retries should be set high enough to allow the cluster to
```
jhecking, line 102:
```suggestion
     * result in errors.  If maxRetries is greater than zero on a write, then
     * sleepBetweenRetries should be set high enough to allow the cluster to
```

lib/policies/batch_apply_policy.js
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// *****************************************************************************
+
+'use strict'
+
+/**
+ * A policy affecting the behavior of batch apply operations.
+ *
+ * @since v6.0.0
```
jhecking, line 22:
```suggestion
 * @since v4.1.0
```

lib/policies/batch_read_policy.js
```diff
@@ -0,0 +1,58 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// *****************************************************************************
+
+'use strict'
+
+/**
+ * A policy affecting the behavior of batch read operations.
+ *
+ * @since v6.0.0
```
jhecking, line 22:
```suggestion
 * @since v4.1.0
```

lib/policies/batch_remove_policy.js
```diff
@@ -0,0 +1,84 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// *****************************************************************************
+
+'use strict'
+
+/**
+ * A policy affecting the behavior of batch remove operations.
+ *
+ * @since v6.0.0
```
jhecking, line 22:
```suggestion
 * @since v4.1.0
```

lib/policies/batch_write_policy.js
```diff
@@ -0,0 +1,84 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// *****************************************************************************
+
+'use strict'
+
+/**
+ * A policy affecting the behavior of batch write operations.
+ *
+ * @since v6.0.0
```
jhecking, line 22:
```suggestion
 * @since v4.1.0
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "aerospike",
-  "version": "4.0.1",
+  "version": "5.0.0",
```
jhecking, line 3: 4.1.0 or 5.0.0? Are there any backwards incompatible changes that warrant a major version jump? I thought this PR only adds new features, so a minor version jump should be sufficient.

test/client.js
```diff
@@ -247,20 +247,21 @@ describe('Client', function () {
...
-      const test = async function (Aerospike, config) {
-        Object.assign(config, { minConnsPerNode: 5, log: { level: 0 } })
-        const client = await Aerospike.connect(config)
-        // give client enough time to establish the requested min. connections
-        await new Promise((resolve) => setTimeout(resolve, 10))
-        const stats = client.stats()
-        client.close()
-        return stats
-      }
+  // context('minConnsPerNode', function () {
```
jhecking, line 250: If no longer needed, just delete the test, rather than just commenting it out?

jhecking (approved):
I think the docs for the new batchWrite/Apply/Delete commands could be improved. Otherwise looks good.

Also, version 4.1 or 5.0?

lib/policies/base_policy.js
rpandian-spike: isn't that used in underlying c-client to "how to retry"?

package.json
rpandian-spike: major version needs to be bumped up because of feature drop - "PredicateExpression filter Replaced by new Expressions Filter".

lib/client.js
```diff
@@ -391,6 +394,254 @@ Client.prototype.batchRead = function (records, policy, callback) {
...
+ *   if (error) throw error
+ *   client.batchApply(batchRecords, udf, function (error, results) {
+ *     if (error) throw error
+ *     results.forEach(function (result) {
+ *       console.log(result)
+ *     })
+ *   })
+ * })
+ */
+Client.prototype.batchApply = function (records, udf, batchPolicy, batchApplyPolicy, callback) {
```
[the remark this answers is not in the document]
rpandian-spike: Fixed the example. batchWrite cannot include batchApply

PR #479, opened 2022-06-11
New workflow for Windows

realmgic:


scripts/build-package.ps1
```diff
@@ -24,6 +24,15 @@ function build_nodejs_client {
...
 # nvm root .\nvm
 
+pwd
+
+cd $Env:USERPROFILE;
+Invoke-WebRequest https://raw.githubusercontent.com/jchip/nvm/v1.5.4/install.ps1 -OutFile install.ps1;
+.\install.ps1 -nvmhome $Env:USERPROFILE\nvm;
+del install.ps1
+
+cd D:\a\aerospike-client-nodejs\aerospike-client-nodejs
```
rpandian-spike, line 34: this is run within aerospike-client-nodejs work dir so no need of this hard coded cd.

rpandian-spike: this is run within aerospike-client-nodejs project on already setup windows so we done need these setup

realmgic: I don't think we use this script in the new workflow - I gave up on the manual `nvm` install and went with the matrix approach.

PR #480, opened 2022-06-16
PR: Client-nodejs: identify policy arg and use it accordingly

rpandian-spike:


rpandian-spike:
this is to address: 
https://aerospike.atlassian.net/browse/CLIENT-1763

lib/client.js
```diff
@@ -1311,16 +1311,33 @@ Client.prototype.info = function (request, host, policy, callback) {
  * })
  */
 Client.prototype.infoAny = function (request, policy, callback) {
+  let req = request
+  let pol = policy
+
   if (typeof request === 'function') {
     callback = request
     request = null
-    policy = null
```
jhecking, line 1317: If the first parameter is a function, then that must be the callback function. Any parameters passed _after_ the callback function should be ignore, or – better yet – a TypeError should be raised.

jhecking (commented):
I don't think it's a good idea to rely on the "type" of the passed in object, by checking the name of the constructor function. The client has never required that policies are passed as instances of a specific policy class, but has instead always accepted plain objects as well, as long as they have the expected attributes. I.e. the client has largely relied on [duck typing](https://en.wikipedia.org/wiki/Duck_typing).

Moving away from Duck Typing and adopting more strict typing like e.g. TypeScript provides might be the right choice. But that shouldn't be done in a piecemeal fashion one function at a time IMHO, but uniformly throughout the client.

lib/client.js
rpandian-spike: A required parameter cannot follow an optional parameter but I see many places we use that way Is that allowed in nodejs?

rpandian-spike: never mind, i see only in couple places like that

jhecking: Callback function is optional. Which examples do you see where a required param follows an optional one?

rpandian-spike:
its as per design, no change needed

PR #481, opened 2022-06-21
PR: Client-NodeJS typescript support

rpandian-spike:
Generated typescript declaration using:
npx -p typescript tsc --declaration --allowJs --emitDeclarationOnly --stripInternal --noUnusedLocals --noUnusedParameters --outFile typings/index
And verified that there is no build error:
tsc examples/test-ts.ts > typings/build_errors.txt

### DIFF
--- a/lib/batch_type.js
+++ b/lib/batch_type.js
@@ -28,27 +28,28 @@ const batchType = as.batchTypes
 // ========================================================================
 // Constants
 // ========================================================================
+exports.types = {
+  /**
+     * An instance of class Record that is used in a batch for read operations.
+     * @const {number}
+     */
+  BATCH_READ: batchType.BATCH_READ,
 
-/**
- * An instance of class Record that is used in a batch for read operations.
- * @const {number}
- */
-exports.BATCH_READ = batchType.BATCH_READ
+  /**
+     * An instance of class Record that is used in a batch for write operations.
+     * @const {number}
+     */
+  BATCH_WRITE: batchType.BATCH_WRITE,
 
-/**
- * An instance of class Record that is used in a batch for write operations.
- * @const {number}
- */
-exports.BATCH_WRITE = batchType.BATCH_WRITE
+  /**
+     * An instance of class Record that is used in a batch for applying record.
+     * @const {number}
+     */
+  BATCH_APPLY: batchType.BATCH_APPLY,
 
-/**
- * An instance of class Record that is used in a batch for applying record.
- * @const {number}
- */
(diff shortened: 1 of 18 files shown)


lib/client.js
```diff
@@ -347,8 +347,8 @@ Client.prototype.batchGet = function (keys, policy, callback) {
  * the batch. This method requires server >= 3.6.0.
  *
  * @param {object[]} records - {@link Record} List of keys and bins to retrieve.
- * @param {type} records[].type - {@link Record#type} Batch type.
- * @param {Key} records[].key - Record Key.
+ * @param {number} records[].type - {@link Record#type} Batch type.
+ * @param {string} records[].key - Record Key.
```
jhecking, line 351: I'm pretty sure this parameter should actually be an instance of `Key` and not just a string. Same for the other batch methods.

lib/batch_type.js
```diff
@@ -28,27 +28,28 @@ const batchType = as.batchTypes
...
- * @const {number}
- */
-exports.BATCH_APPLY = batchType.BATCH_APPLY
-
-/**
- * An instance of class Record that is used in a batch for removal operations.
- * @const {number}
- */
-exports.BATCH_REMOVE = batchType.BATCH_REMOVE
+exports.types = {
```
jhecking, line 31: As discussed, maybe try out `exports = { ... }` to see if you can make this work with TypeScript without breaking the user-facing interface.

lib/query.js
```diff
@@ -417,7 +422,7 @@ Query.prototype.foreach = function (policy, dataCb, errorCb, endCb) {
  *
  * @param {QueryPolicy} [policy] - The Query Policy to use for this operation.
  *
- * @returns {Promise<Record[]>}
+ * @returns {Promise<object[]>}
```
jhecking, line 425: Maybe this is necessary to make the TypeScript type definitions work, but I think when it comes to generating the API docs for the client, just specifying a generic `object` type for the return type is worse.

jhecking (approved):

lib/batch_type.js
rpandian-spike: Updated.

bit0r1n:
Has anyone tested the types for working? For example, I can't initialize the client, because according to the idea of the JS code, I have to provide an object / interface, but in types, I have to give an object of the Config class. Or even enum types are designated as prime numbers (eg Aerospike.(indexType/indexDataType/ttl and other enums). I haven't looked any further as I can't write anything else :)

PR #484, opened 2022-07-28
Update test matrix according to the Readme

eagleeye:
Update test matrix according to the Readme
>The client is compatible with Node.js 10, 12 (LTS), 14 (LTS), 16 (LTS) and 18 (LTS).
https://github.com/aerospike/aerospike-client-nodejs/blob/master/README.md#prerequisites

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: node_js
 node_js:
-- '15'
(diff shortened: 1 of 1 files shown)


eagleeye:
it looks like travis is not used anymore?

PR #489, opened 2022-10-26
[Client-1946] node.js client version 5.1.0

DomPeliniAerospike:
- Improvements
  - [CLIENT-1942] Improved examples within the API documentation.
  - [CLIENT-1943] Added documentation for list, map, hll, and bit expressions.
  - [CLIENT-1944] Revamped existing API documentation.

[CLIENT-1942]: https://aerospike.atlassian.net/browse/CLIENT-1942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLIENT-1943]: https://aerospike.atlassian.net/browse/CLIENT-1943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLIENT-1944]: https://aerospike.atlassian.net/browse/CLIENT-1944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juliannguyen4 (approved):

lib/client.js
```diff
@@ -247,31 +257,46 @@ Client.prototype.removeSeedHost = function (hostname, port) {
...
  *   new Key('test', 'demo', 'key3')
  * ]
- *
- * client.batchExists(keys, function (error, results) {
- *   if (error) {
- *     // handle failure
- *   } else {
- *     results.forEach(function (result) {
+ * 
+ * Aerospike.connect(config, (error, client) => {
```
jhecking, line 278: Just a suggestion: If you are anyway updating all the examples, then it would be better to replace all usage of callback functions with async/await, rather than just replacing regular functions with arrow functions.

docs/overview.md
```diff
@@ -34,25 +33,24 @@ The main modules included in the `aerospike` package are:
...
-  Client class commands such as put, get or query can be sent to an Aerospike
-  database cluster.
+* {@link Client} - The main interface of the Aerospike client. Commands such as put, get or query can be sent to an Aerospike
+  database cluster using the Client class.
 * {@link Key} - Keys are used to uniquely identify a record in the Aerospike database.
 * {@link Record} - Records consists of one or more record "bins" (name-value
-  pairs) and meta-data, incl. time-to-live and generation; a
+  pairs) and meta-data (time-to-live and generation); a
   record is uniquely identified by it's key within a given namespace.
 * {@link Query} - The Query class can be used to perform value-based searches
```
rbotzer, line 50: I think that in the Node.js client Scan should be deprecated and Query handles both primary index (PI) and secondary index (SI) queries now (that's what all language clients should have). If so, please be explicit about it in this description here

rbotzer (approved):

docs/overview.md
DomPeliniAerospike: The functionality may be available already, I will need to take a deeper dive.  The Scan and Query pages still include bits about predicate filtering and the query page doesn't mention searching primary indexes.  I will need to review and reformat those sections in a future release.

PR #490, opened 2022-11-01
Update typings

bit0r1n:
I have created [first version](https://github.com/aerospike/aerospike-client-nodejs/pull/446) of typings, but it was created with errors, which I haven't seen for some unknown reasons (but my projects are still building with my previous typings 😳)
Recently I have decided to update aerospike package in my application, but new typings causes some errors.
For example:
1. I can't initialize client with only "hosts" property, I need to provide all properties, that exists in config interface
Code_UyuxGedWSI
2. I can't normally declare property of my class with "Aerospike Client" - "Client" refers to value (??)
Code_pVWQ9lggXY
And I don't know if there are more problems with current typings, since I can't even initialize client.
Recently @realmgic have pointed me why my typings was removed (btw I have created fork only for using in my project, but it just so happened that I decided to fix my typings). Also I hope, that I have added typings for all new methods, properties, etc.
~~also there are a lot of "any" in current typings 😢~~

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,1563 +1,1947 @@
-declare module "status" {
-    export const ERR_INVALID_NODE: any;
-    export const ERR_NO_MORE_CONNECTIONS: any;
-    export const ERR_ASYNC_CONNECTION: any;
-    export const ERR_CLIENT_ABORT: any;
-    export const ERR_INVALID_HOST: any;
-    export const NO_MORE_RECORDS: any;
-    export const ERR_PARAM: any;
-    export const ERR_CLIENT: any;
-    export const OK: any;
-    export const ERR_SERVER: any;
-    export const ERR_RECORD_NOT_FOUND: any;
-    export const ERR_RECORD_GENERATION: any;
-    export const ERR_REQUEST_INVALID: any;
-    export const ERR_RECORD_EXISTS: any;
-    export const ERR_BIN_EXISTS: any;
-    export const ERR_CLUSTER_CHANGE: any;
-    export const ERR_SERVER_FULL: any;
-    export const ERR_TIMEOUT: any;
-    export const ERR_ALWAYS_FORBIDDEN: any;
-    export const ERR_CLUSTER: any;
-    export const ERR_BIN_INCOMPATIBLE_TYPE: any;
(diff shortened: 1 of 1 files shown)


realmgic:
_I'm commenting here as a user/community contributor, and not as an Aerospike employee._

The fact that you redid the entire typing file (instead of changing/fixing the latest version), so there is no way to see what was changed is very difficult. I tried to see what changed, but this is a completely different file so I can't even tell if it won't break things that already been working.

I think that not having any kind of testing to the typing script is major part of the problem. The client itself has some unit tests (although not for everything, and not perfect) however, the type script does not. There is no way to know it's condition until someone stumble on one of it's issues. 

What about adding some tests, similar to what we have in the regular non-type-script, to make sure everything is working?

bit0r1n:
I think that "big amounts" of changes are visible with current typings. I have fixed my first version of typings, so viewing diff between my first typings and typings from this PR will show not that big amount of changes. As test of typings can be added TS file with calling of methods, creating classes, etc., and run tsc to check if typings are alright.

DomPeliniAerospike:
Hi @bit0r1n,
Sorry for the late response, I started recently, and I should be more responsive in the future.

This looks much different, but if it works for you, this should be fine.

For your convenience and ours, I will also be including file named autoIndex.d.ts to be used as a reference when adding new functions into the index.d.ts file.  I will move the 5.1.1, 5.2.0, and 5.21 functionality to index.d.ts in the next patch.

As Zohar says above, we don't have any unit tests for typescript to know that these work, but I understand you went into extreme detail. Thanks for contributing the Aerospike Nodejs Client.

DomPeliniAerospike:
@bit0r1n
Through talking with my team, I realized that many customers are already using your original description file.  Unfortunately, we will not be able to accept your PR with accompanying unit tests.  If you would like to submit some unit test, you are welcome too, but we can't afford to move ahead with this change unless we are sure it is not a breaking change.

I'm sorry I prematurely accepted this branch, but I hope to add more robust support for typescript in future.

bit0r1n:
I think problem with tests for typings can be solved by rewriting tests to typescript, so it will test typings (compilation before executing test code) and test code behaviour at the "same" time

PR #492, opened 2022-11-11
PR: nodeJS-Client arm64 build support

rpandian-spike:


jhecking:
Looks good to me. But none of the Linux & Windows Client Tests pass. Have you looked into the reason for that?

PR #544, opened 2023-03-27
Nodejs-release-5.3.0

DomPeliniAerospike:


src/main/expressions.cc
```diff
@@ -167,9 +167,8 @@ int convert_entry(Local<Object> entry_obj, as_exp_entry *entry,
 	}
 
 	if (Nan::Has(entry_obj, Nan::New("ctx").ToLocalChecked()).FromJust()) {
-		entry->v.ctx = NULL;
-		rc =
-			get_optional_cdt_context(entry->v.ctx, NULL, entry_obj, "ctx", log);
+		entry->v.ctx =
+			get_optional_cdt_context_heap(&rc, entry_obj, "ctx", log);
```
justinlee-aerospike, line 171: I'm not sure we need to have context stored in the heap?  We can maybe address this later to see if this can be resolved by passing through the stack.

justinlee-aerospike (approved):
Code generally looks okay.  I had a question about the need for heap allocation, but let's get the fix out and maybe we can address this if we get performance enhancement requests in this area.

artanderson (approved):
LGTM. Glad to see the list and map expression params order fixed!

PR #550, opened 2023-03-30
Nodejs-Release-5.4.0

DomPeliniAerospike:


justinlee-aerospike (dismissed):
changes look okay.  I will approve but you should check with Product Management to verify that the current Node.js behavior to have only ordered maps is acceptable.

lib/maps.js
```diff
@@ -324,8 +324,11 @@ exports.writeFlags = as.maps.writeFlags
  * range read.
  * @property {number} VALUE - Return value for single key read and value list
  * for range read.
- * @property {number} KEY_VALUE - Return map items keys and values as an Array,
+ * @property {number} KEY_VALUE - Return map items keys and values as an Array.
  * i.e. [key1, value1, key2, value2, ...].
+ * @property {number} EXISTS - Return true if count > 0.
+ * @property {number} UNORDERED_MAP - Return an ordered map.
+ * @property {number} ORDERED_MAP - Return an unordered map.
```
artanderson, line 331: should these be flipped like below?

```suggestion
 * @property {number} UNORDERED_MAP - Return an unordered map.
 * @property {number} ORDERED_MAP - Return an ordered map.
```

artanderson (dismissed):
Only have the one comment, otherwise looks good

PR #556, opened 2023-05-19
CLIENT-2108

DomPeliniAerospike:


lib/scan.js
```diff
@@ -157,8 +175,147 @@ function Scan (client, ns, set, options) {
...
+   *
+   * ;(async function() {
+   *   try {
+   *     client = await Aerospike.connect(config)
+   *     await client.truncate('test', 'demo', 0)
+   *     await client.batchWrite(batchRecords, {socketTimeout : 0, totalTimeout : 0})
+   *
+   *     const scan = client.scan('test', 'demo', {paginate: true})
+   *     let allResults = []
+   *     let results = await scan.results({maxRecords: 11})
```
justinlee-aerospike, line 262: should this be 10?  or do you need to set maxRecords: 11 if you want a page size of 10?  I'm inferring you wanted a page size of 10 because you said in the comment below that after the three calls, we should have 30 records.

justinlee-aerospike: never mind, I understand now that you are purposefully trying to have an example that will return a partial page as the last page.

justinlee-aerospike (approved):
looks good to me.  would suggest having Brian take a look as well.

artanderson (approved):
LGTM

PR #571, opened 2023-08-02
Nodejs Release 5.7.0

DomPeliniAerospike:


test/query.js
```diff
@@ -902,14 +933,26 @@ describe('Queries', function () {
 
     it('should perform a background query that executes the operations #slow', async function () {
       const query = client.query(helper.namespace, testSet)
-      const ops = [op.incr('backgroundOps', 1)]
+      const ops = [op.write('backgroundOps', 1)]
```
realmgic, line 936: Why?

test/query.js
```diff
@@ -902,14 +933,26 @@ describe('Queries', function () {
...
       await job.waitUntilDone()
 
       const key = keys[Math.floor(Math.random() * keys.length)]
       const record = await client.get(key)
       expect(record.bins.backgroundOps).to.equal(1)
     })
+
+    it('should set TTL to the specified value #slow', async function () {
+      const query = client.query(helper.namespace, testSet)
+      query.ttl = 45
```
realmgic, line 947: i know it's just a test, but people use this for examples and 45 seconds sounds a bit harsh, maybe using 1h?

realmgic (approved):
I think we need more examples for the new stuff - especially for the admin parts - but all in all, good job.

artanderson (dismissed):

lib/policies/admin_policy.js
```diff
@@ -0,0 +1,47 @@
...
+  constructor (props) {
+    props = props || {}
+
+    /**
+     * Maximum time in milliseconds to wait for the operation to complete.
+     *
+     * @type number
+     */
+    this.timeout = props.timeout
+    if (this.timeout === undefined) {
```
bit0r1n, line 41: This check is meaningless, since if `props.timeout` is undefined, then `this.timeout` will be undefined, and this check triggers, and sets it to undefined anyway

lib/query.js
```diff
@@ -392,9 +392,27 @@ function Query (client, ns, set, options) {
...
+   * -1 (defined Aerospike.ttl.NEVER_EXIRE), which means that the record
+   * will get an internal "void_time" of zero, and thus will never expire.
+   *
+   * -2 (defined Aerospike.ttl.DONT_UPDATE), which means that the record
+   * ttl will not change when the record is updated.
+   *
+   * Note that the TTL value will be employed ONLY on background query writes.
+   * @member {number} Query#ttl
+   */
+  this.ttl = 0
```
bit0r1n, line 415: Is there a missing field `.ttl` in options? Same for Scan

test/query.js
DomPeliniAerospike: Makes sense to me, I will add this in before releasing.

test/query.js
DomPeliniAerospike: The same bin is incremented in other tests.  Writing here instead of incrementing allows the result in the bin to be a consistent value rather than dependent on the execution of other tests.

Here is the test failure for more information:

`  1) Queries
       query.operate()
         should perform a background query that executes the operations #slow:

      AssertionError: expected 3 to equal 1
      + expected - actual

      -3
      +1
      
      at Context.<anonymous> (test/query.js:942:44)

`

lib/policies/admin_policy.js
DomPeliniAerospike: This is correct.  This same useless check was present in info_policy.js.  I removed it from both files now.  Thank you.

lib/query.js
DomPeliniAerospike: Just added changes to allow the TTL to be set in query and scan options as well.  Added tests to verify correct behavior on scans and queries. This was a nice find.  Thank you!

mcoberly2 (approved):

PR #607, opened 2024-02-14
Nodejs-Release-5.11.0

DomPeliniAerospike:


lib/lists.js
```diff
@@ -1079,6 +1079,61 @@ exports.clear = function (bin) {
   return new ListOperation(opcodes.LIST_CLEAR, bin)
 }
 
+/**
+ * @summary Creates map create operation.
```
artanderson, line 1083:
```suggestion
 * @summary Creates list create operation.
```

artanderson (approved):
Just one small comment, otherwise LGTM

realmgic (approved):

lib/lists.js
DomPeliniAerospike: Corrected this in the most recent commit.

PR #630, opened 2024-08-07
Fixing signature of `batchRead` (should accept array of records inste…

atychyna:
…ad of single record)


DomPeliniAerospike:
This is fixed in version 5.13.1.

```
public batchRead(records: BatchReadRecord[], policy?: policy.BatchPolicy): Promise<BatchResult[]>;
```

PR #697, opened 2025-07-18
Metrics V2 and Dynamic Config

DomPeliniAerospike:
CLIENT-3407
CLIENT-3408


src/main/util/conversions.cc
```diff
@@ -369,6 +369,32 @@ int get_optional_uint16_property(uint16_t *intp, bool *defined,
...
+								 Local<Object> obj, char const *prop,
+								 const LogInfo *log)
+{
+	Nan::HandleScope scope;
+	Local<Value> value =
+		Nan::Get(obj, Nan::New(prop).ToLocalChecked()).ToLocalChecked();
+	if (value->IsNumber()) {
+		if (defined != NULL)
+			(*defined) = true;
+		(*intp) = Nan::To<uint32_t>(value).FromJust();
```
AerospikeNate-L, line 382: assigns only the least significant byte of a 4-byte unsigned int to the `uint8_t` that `intp` points to..
should `uint32_t` be `uint8_t` here fore type-safety? maybe copy-paste from `get_optional_uint32_property` that didnt get refactored?

typings/index.d.ts
```diff
@@ -16771,6 +16891,14 @@ export namespace filter {
...
  */
 declare namespace statusNamespace {
+    /**
+     * There is a conflict between metrics enable/disable and dynamic configuration metrics.
+     */
+    export const AEROSPIKE_METRICS_CONFLICT = -19;
+    /**
+     * There is a conflict between metrics enable/disable and dynamic configuration metrics.
+     */
+    export const METRICS_CONFLICT = -19;
```
AerospikeNate-L, line 16901: is `AEROSPIKE_METRICS_CONFLICT` -19 or -20?

src/main/policy.cc
```diff
@@ -720,22 +720,60 @@ int adminpolicy_from_jsobject(as_policy_admin *policy, Local<Object> obj,
...
 {
 	if (obj->IsUndefined() || obj->IsNull()) {
 		return AS_NODE_PARAM_ERR;
 	}
 	int rc = 0;
 	as_metrics_policy_init(policy);
 
+	bool defined = false;
+
+	int report_dir_size = 256;
```
AerospikeNate-L, line 733: is this the max path string length? do we have this documented somewhere?

AerospikeNate-L (commented):
probably dont know enough about nodejs / ts to give this an exhaustive review, but a few little nits / confusion poinst stuck out

Otherwise if tests pass & examples work, looks good.

typings/index.d.ts
DomPeliniAerospike: Will correct this to -20.

src/main/util/conversions.cc
DomPeliniAerospike: This is standard practice for the Nodejs Client.  The max number is 2^53 in nodejs. We don't size check for performance reasons, so this truncation practice is standard.

The conversion function `get_optional_uint16_property` uses `Nan::To<uint32_t>` to convert values as well. 
`Nan::To<uint32_t>` is the smallest value available for this function, but since we are truncating anyways, we can still use the helper function and just truncate the value into a smaller value.

We also only use this value for small integers in non-database operations, so the result of a truncation should be relatively inconsequential. 

Happy to talk more about this if you are still concerned.

PR #725, opened 2025-08-07
typings: missing generics for batch methods (where record's bins can exist)

bit0r1n:
And fix BatchResult.record type


DomPeliniAerospike:
Added this in a new commit to fix merge conflicts: https://github.com/aerospike/aerospike-client-nodejs/pull/739.

Thank you for contributing @bit0r1n!

PR #727, opened 2025-08-28
CLIENT-3587: Added support for user agent

DomPeliniAerospike:


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "aerospike-client-c"]
 	path = aerospike-client-c
 	url = https://github.com/aerospike/aerospike-client-c.git
-	branch = stage
+	branch = CLIENT-3585-Add-Global-To-Header
```
juliannguyen4, line 4: Hey Dominic, about this C client change:

https://github.com/aerospike/aerospike-client-c/commit/88252b538b830a208c331caf1f8dd2a418e232ee

Could you clarify why you added declarations for `aerospike_client_language` and `aerospike_client_version`?

DomPeliniAerospike: Resolved this offline.

PR #737, opened 2025-10-06
Add npm dependency scanning to dependabot

DomPeliniAerospike:
Expands dependabot to monitor all package.json files for security vulnerabilities


DomPeliniAerospike (commented):
Shouldn't we be using cooldown to make sure we aren't updating these dependencies as soon as updates come out?  The original guidance was 2 days, but I think a week cooldown before pulling any changes it probably the most safe.

Otherwise , LGTM!

PR #739, opened 2025-10-07
Added generic typing to batch functions

DomPeliniAerospike:
Contributor: @bit0r1n


typings/index.d.ts
```diff
@@ -295,7 +295,7 @@ export class AerospikeRecord<B extends AerospikeBins = AerospikeBins> {
      *
      * @type {AerospikeBins}
      */
-    public bins: B;
+    public bins: AerospikeRecord<B>;
```
bit0r1n: This change was moved incorrectly. It was originally made for `BatchResult` class and `.record` field, but now it affects `AerospikeRecord` class, and for now, it's impossible to access record's bins, since there is loop reference to AerospikeRecord type, instead of bins type (`B` generic)

PR #799, opened 2025-12-02
fix(docs): incompatible changes

jefflembeck:
Support was added for Node 24 in 6.3 and according to release docs, nothing was dropped.

In 6.4.0, support was removed for RHEL 8 and Debian 11

This takes care of both of those

### DIFF
--- a/incompatible.md
+++ b/incompatible.md
@@ -2,8 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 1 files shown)


DomPeliniAerospike:
Added the fix in this [PR](https://github.com/aerospike/aerospike-client-nodejs/pull/814).

Thanks for point this out!

PR #800, opened 2025-12-02
fix(typings): AerospikeRecord

### LINKED ISSUE #801: AerospikeRecord.bins incorrectly typed as AerospikeRecord<B> instead of B
**Describe the bug**
The bins property on AerospikeRecord in typings/index.d.ts is incorrectly typed as AerospikeRecord<B> instead of B. This prevents indexing bins with a string key, which is the (at least it is our) primary way bins are accessed.

**To Reproduce**
```typescript
import Aerospike from 'aerospike';

const client = await Aerospike.connect(config);
const key = new Aerospike.Key('namespace', 'set', 'myKey');
const record = await client.get(key);

// TypeScript error: Element implicitly has an 'any' type because expression 
// of type 'string' can't be used to index type 'AerospikeRecord<AerospikeBins>'.
const value = record.bins['myBin'];
```

Error message:

```
TS7053: Element implicitly has an 'any' type because expression of type 'string' 
can't be used to index type 'AerospikeRecord<AerospikeBins>'.
  No index signature with a parameter of type 'string' was found on type 'AerospikeRecord<AerospikeBins>'.
```

**Expected behavior**
`record.bins` should be typed as `B` (which extends `AerospikeBins = { [key: string]: any }`), allowing string indexing to access bin values.

**Versions and Environment (please complete the following information):**
- Client application OS: macOS 15.2
- Aerospike Client Version: 6.4.0
- Aerospike Database Version: N/A

jefflembeck:
This should be B, as it extends AerospikeBins, not AerospikeRecord<B>, which would lead to a nested record.

(This change reflects the JSDoc comment)

Fixes #801

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -295,7 +295,7 @@ export class AerospikeRecord<B extends AerospikeBins = AerospikeBins> {
      *
      * @type {AerospikeBins}
      */
-    public bins: AerospikeRecord<B>;
+    public bins: B;
 
     /**
      * The record's remaining time-to-live in seconds before it expires.


DomPeliniAerospike:
Made a fix in this [PR](https://github.com/aerospike/aerospike-client-nodejs/pull/812).

Thanks for the fix!

PR #803, opened 2025-12-02
fix(typings): type BatchResult.record as Record

### LINKED ISSUE #802: BatchResult.record incorrectly typed as B instead of AerospikeRecord<B>
**Describe the bug**
The record property on BatchResult in typings/index.d.ts is typed as B (the bins type) instead of AerospikeRecord<B>. This contradicts the actual runtime behavior and causes type errors when accessing result.record.bins.

**To Reproduce**
```typescript
import Aerospike, { AerospikeBins } from 'aerospike';

const client = await Aerospike.connect(config);

const batchKeys = keys.map((key) => ({
    type: Aerospike.batchType.BATCH_READ,
    key: new Aerospike.Key(namespace, set, key),
    readAllBins: true,
}));

const results = await client.batchRead(batchKeys);

// TypeScript error when accessing record.bins
const bins = results
    .map((row) => row.record.bins)
    .filter((c: AerospikeBins | undefined) => !!c);
```

Error message:

```
TS2769: No overload matches this call.
  Overload 1 of 2, '(predicate: (value: AerospikeBinValue, ...) => value is AerospikeBins, ...)', gave the following error.
    Types of parameters 'c' and 'value' are incompatible.
      Type 'AerospikeBinValue' is not assignable to type 'AerospikeBins | undefined'.
        Type 'null' is not assignable to type 'AerospikeBins | undefined'.
```

The underlying issue is that `row.record` is typed as `AerospikeBins`, so `row.record.bins` returns `AerospikeBinValue` (via the string index signature) instead of the actual bins object.

**Expected behavior**
BatchResult.record should be typed as `AerospikeRecord<B>`, allowing `result.record.bins` to return `B` (the bins object).

The runtime implementation in [lib/commands/batch_command.js](https://github.com/aerospike/aerospike-client-nodejs/blob/master/lib/commands/batch_command.js#L36) confirms this:

```javascript
const record = new Record(result.key, result.bins, result.meta)
return new BatchResult(result.status, record, result.inDoubt)
```

At runtime, record is a Record instance (i.e., AerospikeRecord) constructed with key, bins, and meta — not just the bins object directly.

**Versions and Environment (please complete the following information):**
 - Client application OS: MacOS 15
 - Aerospike Client Version: 6.3.0/6.4.0
 - Aerospike Database Version: N/A

jefflembeck:
Fixes #802

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -823,7 +823,7 @@ export class BatchResult<B extends AerospikeBins = AerospikeBins> {
     /**
      * Construct a new BatchResult instance.
      */
-    public constructor(status: typeof statusNamespace[keyof typeof statusNamespace], record: B, inDoubt: boolean);
+    public constructor(status: typeof statusNamespace[keyof typeof statusNamespace], record: AerospikeRecord<B>, inDoubt: boolean);
     /**
      * Result code for this returned record. If not {@link statusNamespace.AEROSPIKE_OK|AEROSPIKE_OK}, the record will be null.
      */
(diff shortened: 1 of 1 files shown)


DomPeliniAerospike:
Made fix for this in this [PR](https://github.com/aerospike/aerospike-client-nodejs/pull/812).

Thanks the fix!

PR #881, opened 2025-12-14
typings: fix Nodejs internal types resolving

bit0r1n:
Since this project is not written on TS and don't have tsconfig file, which describes that this project can use Nodejs types - there are only abstract types available for `.d.ts` files, so no `Buffer`, `EventEmitter` and other Nodejs-only types.

Because of this limited list of types, there are some errors with typings - `Client` class extends `any` instead of `EventEmitter`, same with `RecordStream` and `Stream`; `PartialAerospikeBinValue` unions with `Buffer`, which is only available on Nodejs, this results in `type AerospikeBinValue = any` or even can cause an error with typescript-eslint.

It can be fixed with `reference` tag in `.d.ts` file and without tsconfig file.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,4 +1,4 @@
-import * as Buffer from "buffer";
+/// <reference types="node" />
 import { EventEmitter, Stream } from "stream";
 
 /**


DomPeliniAerospike:
HI @bit0r1n,
I will add this fix into the next release.
I will also lint with typescript-eslint in order to avoid this some of this in the future.
If you have any more suggestions or thoughts for improving typescript support, I would appreciate any feedback!
Thanks!

bit0r1n:
> If you have any more suggestions or thoughts for improving typescript support, I would appreciate any feedback!

It would be good to re-write `lib` part of repo to TypeScript, so that there is no need to conditionally maintain two pieces of code, when this could be combined by maintaining a project in TypeScript. Also probably with that, project can use modern syntax instead of deprecated methods (for example, there are a lot of class definitions with `function`s and `util.inherits`, it can be replaced with `class` and `extends` keywords, so it would be more readable)

But if that will be too complicated, it also would be nice to split `index.d.ts` into multiple files to follow structure as in `lib` folder, so it would be easier to work with in future.

DomPeliniAerospike:
In the Next generation Aerospike Node.js Client, we are strongly considering re-writing the project in typescript.  Since doing so will take considerable time, we can work on restructuring the typescript file structure to match `lib`.

Thanks for the suggestion!

DomPeliniAerospike:
Hi @bit0r1n,
This fix was added in [6.5.2](https://github.com/aerospike/aerospike-client-nodejs/releases/tag/v6.5.2)

PR #920, opened 2026-01-21
CLIENT-3751: Added Path Expressions support

DomPeliniAerospike:
exp.selectByPath
operations.selectByPath
exp.modifyByPath
operations.modifyByPath
loop variables
loopVarPart
pathSelectFlags
pathModifyFlags


lib/cdt_context.js
```diff
@@ -221,6 +224,31 @@ class CdtContext {
     return this.add(MAP_VALUE, value)
   }
 
+  /**
+   * @summary At the current context, causes a query to return a list of all the children
+   * of the current item. For a map, this will recurse into the map elements.
+   * For a list, this will include all the children in the list.
+   *
+   * @param {any} value - Map value
```
juliannguyen4: Doesn't this call take in no parameters?

lib/cdt_context.js
```diff
@@ -221,6 +224,31 @@ class CdtContext {
...
+    return this.add(EXP)
+  }
+
+  /**
+   * @summary All children of the current level will be selected, and then the filter expression
+   * is applied to each item in turn.  Items that cause the expression to evaluate to true will be added to the
+   * list of items returned in a query for this level.  Items that cause the expression to evaluate to false
+   * will be filtered out
+   *
+   * @param {any} value - Map value
```
juliannguyen4: I think this parameter should be `exp`?

lib/exp.js
```diff
@@ -1101,3 +1154,116 @@ exports.expWriteFlags = {
    */
   EVAL_NO_FAIL: writeFlags.EVAL_NO_FAIL
 }
+
+/**
+ *
+ * @readonly
+ * @enum {number}
+ * @description Path expression select operation flags. Use BITWISE OR to combine flags.
```
juliannguyen4, line 1263: > Use BITWISE OR to combine flags

This is good to point out!

lib/exp.js
```diff
@@ -1101,3 +1154,116 @@ exports.expWriteFlags = {
...
+  /**
+   * Return the list of the values of the nodes finally selected by the context.
+   * For maps, this returns the value of each (key, value) pair.
+   * @const {number}
+   */
+  VALUE: pathSelectFlags.VALUE,
+
+  /**
+   * Return the list of the values of the nodes finally selected by the context.
+   * This is a synonym for AS_EXP_PATH_SELECT_VALUE to make it clear in your
```
juliannguyen4: nit: I would reference `VALUE` from the Node.js client instead of the C client's `AS_EXP_PATH_SELECT_VALUE`

lib/exp.js
```diff
@@ -1101,3 +1154,116 @@ exports.expWriteFlags = {
...
+   * source code that you're expecting a list.
+   * @const {number}
+   */
+  LIST_VALUE: pathSelectFlags.LIST_VALUE,
+
+  /**
+   * Return the list of map values of the nodes finally selected by the context.
+   * This is a synonym for AS_EXP_PATH_SELECT_VALUE to make it clear in your
```
juliannguyen4: See comment above

lib/exp.js
```diff
@@ -1101,3 +1154,116 @@ exports.expWriteFlags = {
...
+   * source code that you're expecting a map.  See also
+   * AS_EXP_PATH_SELECT_MAP_KEY_VALUE.
+   * @const {number}
+   */
+  MAP_VALUE: pathSelectFlags.MAP_VALUE,
+
+  /**
+   * Returns the list of map (key, value) pairs of the nodes finally selected
+   * by the context.  This is a synonym for setting both
+   * AS_EXP_PATH_SELECT_MAP_KEY and AS_EXP_PATH_SELECT_MAP_VALUE bits together.
```
juliannguyen4, line 1304: nit: see above comment.

lib/exp.js
```diff
@@ -1101,3 +1154,116 @@ exports.expWriteFlags = {
...
+   * Return the list of map values of the nodes finally selected by the context.
+   * This is a synonym for AS_EXP_PATH_SELECT_VALUE to make it clear in your
+   * source code that you're expecting a map.  See also
+   * AS_EXP_PATH_SELECT_MAP_KEY_VALUE.
+   * @const {number}
+   */
+  MAP_VALUE: pathSelectFlags.MAP_VALUE,
+
+  /**
+   * Returns the list of map (key, value) pairs of the nodes finally selected
```
juliannguyen4, line 1302: Does this docstring belong to `MAP_KEY_VALUE` instead?

Also it would be better to note how the list would be formatted. Like in Python, I added this detail:

```
The list is formatted as [key0, value0, key1, value1...].
```

lib/exp.js
```diff
@@ -1101,3 +1154,116 @@ exports.expWriteFlags = {
...
+ * @readonly
+ * @enum {number}
+ * @description Loop variable parts for expression loop variables.
+ * Used to specify which part of a loop variable to access during iteration.
+ */
+exports.loopVarPart = {
+  /**
+   * Access the key part of the loop variable.
+   * For maps, this refers to the map key.
+   * For lists, this refers to the list index.
```
juliannguyen4, line 1353: For the Python client I only documented that this works for maps

```rst
.. data:: EXP_LOOPVAR_KEY

    The key associated with this value if part of a key-value pair of a map.
```

To get list indices, you could use `INDEX` already right?

src/main/enums/exp_path_modify_flags.cc
```diff
@@ -0,0 +1,38 @@
...
+
+extern "C" {
+#include <aerospike/as_status.h>
+#include <aerospike/as_operations.h>
+}
+
+using namespace v8;
+
+#define set(__obj, __name, __value)                                            \
+	Nan::Set(__obj, Nan::New(__name).ToLocalChecked(), Nan::New(__value))
```
juliannguyen4, line 28: Noticed this macro is also defined in `src/main/enums/exp_path_select_flags.cc`. Maybe move this to a header that is included by both source files?

src/main/cdt_ctx.cc
```diff
@@ -209,26 +213,42 @@ int get_optional_cdt_context(as_cdt_ctx *context, bool *has_context,
 			as_cdt_ctx_add_map_key_create(context, asValue, (as_map_order) 3);
 			as_v8_detail(log, "Adding Map Value context");
 			break;
+		case (AS_CDT_CTX_EXP):
+			if (v8value->IsArray()) {
+				Local<Array> exp_ary = Local<Array>::Cast(v8value);	
+				as_exp *exp = NULL;
+				if (compile_expression(exp_ary, &exp, log) != AS_NODE_PARAM_OK) {
+					as_v8_error(log, "Expressions could not be compiled");
```
juliannguyen4, line 225: Just for my understanding: does `compile_expression` already set an error using `log`? I'm worried that this line might overwrite a more specific error message reported in `compile_expression`

src/main/cdt_ctx.cc
```diff
@@ -209,26 +213,42 @@ int get_optional_cdt_context(as_cdt_ctx *context, bool *has_context,
...
 		}
+
 	}
 
 	return AS_NODE_PARAM_OK;
 
 }
 
-as_cdt_ctx* get_optional_cdt_context_heap(int* rc,
+as_cdt_ctx* get_cdt_context_heap(int* rc,
```
juliannguyen4, line 248: Just a future long term goal, it might be helpful to combine these helper functions `get_cdt_context_heap` and `get_optional_cdt_context` together since they seem to share a lot of similar code

src/main/expressions.cc
```diff
@@ -198,10 +204,34 @@ int convert_entry(Local<Object> entry_obj, as_exp_entry *entry,
...
+			Local<Array> exp_ary = Local<Array>::Cast(exp_obj);
+			if (compile_expression(exp_ary, &exp, log) != AS_NODE_PARAM_OK) {
+				return AS_NODE_PARAM_ERR;
+			}
+		}
+		else if (exp_obj->IsString()) {
+			Nan::Utf8String exp_b64(exp_obj.As<String>());
+			exp = as_exp_from_base64(*exp_b64);
+		}
+		else{
```
juliannguyen4, line 226: Long term TODO: there's also similar code to convert expressions (either list or base64 string) to an `as_exp` object in other parts of the Node.js glue code; it might be helpful to create a helper function to encapsulate this logic

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+const loopVarPart: any = exp.loopVarPart
+const type: any = exp.type
+
+const keygen = helper.keygen
+const tempBin = 'ExpVar'
+
+describe('Aerospike.exp.selectByPath', async function () {
+  const client = helper.client
+
+  helper.skipUnlessVersion('>= 5.0.0', this)
```
juliannguyen4, line 43: Nit: For the Python client, I only check the server version before running the dev tests. At a specific commit in the python client repo, I treat the tests in that commit as only compatible with the Python client build for that commit.

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+
+  const hll2 = Buffer.from([8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 16, 0, 0, 0, 0,
+            0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 65, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
```
juliannguyen4: Is there a helper function to create a buffer with zeroes times a certain number? It would be easier to read

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+                const modifyByPath = exp.modifyByPath(exp.binMap('floatMap'), exp.type.MAP, modExpression, noFailFlags, addAllChildren)
+  
+                await verifyModifyByPath('floatMap', modifyByPath, {a: 14, b: 14, c: 14})
+              })
+
+            })
+
+            context('NO_FAIL', function () {
+              it('equals the correct numeric value', async function () {
+                expect(pathModifyFlags.NO_FAIL).to.eql(16)
```
juliannguyen4, line 194: I don't believe we need to check the internal value `16` since downstream users are supposed to use `pathModifyFlags.NO_FAIL` directly instead of `16`

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+  context('Positive tests', function () {
+
+
+    context('modifyByPath', function () {
+      context('arguments', function () {
+
+        context('bin', function () {
+
+          it('accepts exp.mapBin', async function () {
+            const ctx: cdt.Context = new Context().addMapKey('book').addAllChildren().addMapKey('price')
```
juliannguyen4: This variable isn't used in this test case.. it made the test case kinda confusing to understand at first

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+
+            context('MATCHING_TREE', function () {
+
+              const flags = pathModifyFlags.DEFAULT
+              const noFailFlags = pathModifyFlags.DEFAULT | pathModifyFlags.NO_FAIL
+
+              it('equals the correct numeric value', async function () {
+                expect(flags).to.eql(0)
+                expect(noFailFlags).to.eql(16)
+              })
```
juliannguyen4, line 172: See other comment above about testing internal values for enums

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+
+        context('flags', function () {
+
+
+          context('pathSelectFlags', function () {
+
+            context('MATCHING_TREE', function () {
+
+              const flags = pathModifyFlags.DEFAULT
+              const noFailFlags = pathModifyFlags.DEFAULT | pathModifyFlags.NO_FAIL
```
juliannguyen4, line 167: Here should specify MATCHING_TREE explicitly?

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+            })
+
+          })
+        })
+
+        context('modExp', function () {
+
+
+          it('modifies with standard expression', async function () {
+            const ctx: cdt.Context = new Context().addMapKey('book').addAllChildren().addMapKey('price')
```
juliannguyen4: `ctx` also not used here. There may be a linter for typescript to report warnings about unused variables

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+
+            const modExpression = exp.float(14.0)
+
+            const modifyByPath = exp.modifyByPath(exp.binMap('c_example'), exp.type.MAP, modExpression, pathModifyFlags.DEFAULT, addAllChildren)
```
juliannguyen4, line 208: Nit: may be helpful to create a variable for the bin name `c_example` so you can more easily refactor / rename later

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+            const modifyByPath = exp.modifyByPath(exp.binList('dnFloatList'), exp.type.LIST, modExpression, pathModifyFlags.DEFAULT, ctx)
+            
+            await verifyModifyByPath('dnFloatList', modifyByPath, [[[ 8.88, 17.76, 26.64 ]]]  )
+
+          })
+
+          context('loopVar', function () {
+
+            it('use loopVarFloat Expression', async function () {
+              const modExpression = exp.cond(exp.gt(exp.loopVarFloat(exp.loopVarPart.VALUE), exp.float(3.6)), exp.float(1.0), exp.float(0.0))
```
juliannguyen4, line 259: For readability it'd be helpful to break each expression into it's own line as well as indent each expression used as an argument

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+
+
+              const modifyByPath = exp.modifyByPath(exp.binList('listList'), exp.type.LIST, modExpression, pathModifyFlags.DEFAULT, addAllChildren)
+              
+              await verifyModifyByPath('listList', modifyByPath, [ false, true, true ]  )
+
+
+            })
+
+            it('use loopVarStr Expression with loopVarPart.KEY', async function () {
```
juliannguyen4, line 400: Nit: Noticed above that contexts can be used to organize test cases; if you could move the test cases for LoopVarPart into their own context, it's more explicit that you're testing LoopVarPart

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+              it('returns the correct value when used with operate', async function () {
+                const selectByPath = exp.selectByPath(exp.binMap('floatMap'), exp.type.MAP, flags, addAllChildren)
+  
+                await verifySelectByPath('floatMap', selectByPath, record.floatMap)             
+              })
+
+              it('returns the correct value when used with operate and NO_FAIL', async function () {
+                const selectByPath = exp.selectByPath(exp.binMap('floatMap'), exp.type.MAP, noFailFlags, addAllChildren)
+  
+                await verifySelectByPath('floatMap', selectByPath, record.floatMap)
```
juliannguyen4, line 525: I think the behavior for NO_FAIL should be tested explicitly to check that selectByPath does not fail when the filter expression iterates on an unexpected type.

ts-test/tests/path_expressions.ts
```diff
@@ -0,0 +1,756 @@
...
+            context('VALUE', function () {
+              const flags = pathSelectFlags.VALUE
+              const noFailFlags = pathSelectFlags.VALUE | pathSelectFlags.NO_FAIL
+
+              it('equals the correct numeric value', async function () {
+                expect(flags).to.eql(1)
+                expect(noFailFlags).to.eql(17)
+              })
+
+              it('returns the correct value when used with operate', async function () {
```
juliannguyen4, line 539: If you could parametrize the test cases it would reduce the number of lines needed to maintain the tests. Noticing there's some lines that are repeated in each case like `exp.selectByPath` and `verifySelectByPath`.

ts-test/tests/path_operations.ts
```diff
@@ -0,0 +1,683 @@
...
+describe('Path Operations', async function () {
+  const client = helper.client
+
+  helper.skipUnlessVersion('>= 5.0.0', this)
+
+  const key: Key = new Aerospike.Key(helper.namespace, helper.set, 1)
+
+  const addAllChildren: cdt.Context = new Context().addAllChildren()
+  const doubleAddAllChildren: cdt.Context = new Context().addAllChildren().addAllChildren()
+
```
juliannguyen4, line 50: Nit: could include these, as well as the definition for `record`, in a helper module shared by both this test file and `tests/path_expressions.ts`

ts-test/tests/path_operations.ts
```diff
@@ -0,0 +1,683 @@
...
+        })
+
+        context('flags', function () {
+
+
+          context('pathSelectFlags', function () {
+
+            context('MATCHING_TREE', function () {
+
+              const flags = pathModifyFlags.DEFAULT
```
juliannguyen4, line 157: I made a similar comment in `tests/path_expressions.ts`. Just noting that a lot of comments from that file can also apply to here

ts-test/tests/path_operations.ts
```diff
@@ -0,0 +1,683 @@
...
+            context('NO_FAIL', function () {
+              it('equals the correct numeric value', async function () {
+                expect(pathModifyFlags.NO_FAIL).to.eql(16)
+              })
+            })
+
+          })
+        })
+
+        context('modExp', function () {
```
juliannguyen4, line 190: Think a more helpful name for this is modifyByPath, since we're testing that specific operation

ts-test/tests/path_operations.ts
```diff
@@ -0,0 +1,683 @@
...
+    })
+  })
+
+
+  context('Negative tests', function () {
+    context('selectByPath', function () {
+      context('arguments', function () {
+        context('bin', function () {
+
+                          //const selectByPath = exp.selectByPath(exp.binMap('floatMap'), exp.type.LIST, noFailFlags, addAllChildren)
```
juliannguyen4: nit: unused var

ts-test/tests/path_operations.ts
```diff
@@ -0,0 +1,683 @@
...
+              assert.fail("An error should have been caught!")
+            }
+            catch(error: any){
+              expect(error.message).to.eql("Operations array invalid")
+            }          
+          })
+        })
+
+        context('context', function () {
+          it('Doe not accept non-number values', async function () {
```
juliannguyen4, line 562: This is testing valueType and not context

ts-test/tests/path_operations.ts
```diff
@@ -0,0 +1,683 @@
...
+            catch(error: any){
+              expect(error.message).to.eql("Operations array invalid")
+            }          
+          })
+        })
+
+        context('pathSelectFlags', function () {
+          it('Doe not accept non-number values', async function () {
+            const ops: operations.Operation[] = [
+              op.selectByPath('floatMap', addAllChildren, 'invalid' as any)
```
juliannguyen4: I think this is missing a 4th required argument at the end?

ts-test/tests/path_operations.ts
```diff
@@ -0,0 +1,683 @@
...
+            catch(error: any){
+              expect(error.message).to.eql("Operations array invalid")
+            }
+          })
+        })
+
+        context('context', function () {
+          it('Doe not accept non-number values', async function () {
+            const ops: operations.Operation[] = [
+              op.modifyByPath('floatMap', null as any, exp.float(14.0), exp.pathModifyFlags.DEFAULT),
```
juliannguyen4: Here is also testing valueType instead of context

typings/index.d.ts
```diff
@@ -1679,15 +1679,33 @@ export namespace cdt {
...
         public addMapKey(key: string): CdtContext;
+         /**
+         * Lookup map by base map's key. If the map at key is not found,
+         * create it with the given sort order at that key.
+         *
+         * @param {any} key - Map key
+         * @param {number} order - Sort order used if a map is created
+         * @return {CdtContext} Updated CDT context, so calls can be chained.
+         */
+        public addMapKeyCreate(key: string, order?: maps.order): CdtContext;
```
juliannguyen4, line 1690: Just for my understanding what's the difference between the docstrings in both `typings/*.ts` and `lib/*.js`?

typings/index.d.ts
```diff
@@ -16547,6 +16564,95 @@ export namespace exp {
...
+
+    /**
+     * @readonly
+     * @remarks Path expression modify operation flags.
+     */
+    export enum loopVarPart {
+        /**
+         * Access the key part of the loop variable.
+         * For maps, this refers to the map key.
+         * For lists, this refers to the list index.
```
juliannguyen4, line 16640: Wrote another comment about loopVarPart.KEY above

typings/index.d.ts
```diff
@@ -17348,6 +17454,110 @@ export namespace exp {
...
+     */
+    export const loopVarHLL: (varId: exp.loopVar) => AerospikeExp;
+
+    /**
+     * Path select expression.
+     * Used in path expressions.
+     *
+     * @param bin - bin name.
+     * @param valueType - expression value type.
+     * @param flags - path selct flags.
```
juliannguyen4: nit: typo

typings/index.d.ts
```diff
@@ -17348,6 +17454,110 @@ export namespace exp {
...
+     */
+    export const selectByPath: (bin: AerospikeExp, valueType: exp.type, flags: exp.pathSelectFlags, ctx: cdt.Context) => AerospikeExp;
+    /**
+     * Path modify expression.
+     * Used in path expressions.
+     *
+     * @param bin - bin name.
+     * @param valueType - expression value type.
+     * @param modExp - expression for modificaiton.
+     * @param flags - path selct flags.
```
juliannguyen4: nit: typo.

typings/index.d.ts
```diff
@@ -17492,6 +17702,27 @@ export namespace operations {
      * @see {@link !ttl} for "special" TTL values.
      */
     export function touch(ttl: number): Operation;
+    /**
+     * Path select operation.
+     * Used in path expressions.
+     *
+     * @param bin - bin name.
+     * @param context - CDT Context.
+     * @param flags - path selct flags.
```
juliannguyen4: typo

binding.gyp
```diff
@@ -80,6 +80,7 @@
         'src/main/map_operations.cc',
         'src/main/bit_operations.cc',
         'src/main/hll_operations.cc',
+        'src/main/cdt_operations.cc',
```
juliannguyen4, line 83: Long term suggestion (unrelated): may be better to recursively get all the source code files. In the Python client I use this line instead so I don't have to update my compile script `setup.py`:
```Python
source_files = glob.glob(pathname="src/main/**/*.c", recursive=True)
```

out.txt
```diff
@@ -0,0 +1,2 @@
+link eth0 state up
```
juliannguyen4: Assuming this was used for debugging. Should be removed?

juliannguyen4 (commented):
Hi Dominic, I reviewed the whole PR and left some suggestions. If you could address them that'd be great.

lib/cdt_context.js
DomPeliniAerospike: Fixed docstring

lib/cdt_context.js
DomPeliniAerospike: Fixed this docstring

lib/exp.js
DomPeliniAerospike: Fixed enum value

lib/exp.js
DomPeliniAerospike: Fixed enum value

lib/exp.js
DomPeliniAerospike: Added this note

lib/exp.js
DomPeliniAerospike: Fixed enum value

lib/exp.js
DomPeliniAerospike: I simply copied the Java client string here, and since the Java Client says it can be used with lists, it should be fine to do so.

src/main/enums/exp_path_modify_flags.cc
DomPeliniAerospike: I will add a ticket for this, would be good to reduce duplicate code.

src/main/cdt_ctx.cc
DomPeliniAerospike: compile_expression does not set a specific error code, only logs an error and returns an error code. No context should be lost.

src/main/cdt_ctx.cc
DomPeliniAerospike: I will add a ticket for combining these functions. duplicate code could be cut back with a simple flag.

src/main/expressions.cc
DomPeliniAerospike: i will add a ticket for a compile expression helper function to cut back duplicate code.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: I will discuss this with you further at a later date.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: Adding a helper function would be good, will add a ticket for this.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: The test simply ensures that the C++ pass through value hasn't changed.  It shouldn't fail if it isn't 16, it should fail if it doesn't match the expected value.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: Changed it to actually use the ctx variable.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: Resolved above

ts-test/tests/path_expressions.ts
DomPeliniAerospike: They should match MATCHING_TREE, but since it is a pathModify operation, we use DEFAULT instead of MATCHING_TREE.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: Removed this usage.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: Added a ticket for improving the readability of these expressions.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: I will add a ticket for shorthand naming and better organization of records.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: Added test case for this.

ts-test/tests/path_expressions.ts
DomPeliniAerospike: Test could be parameterized further.  Making a ticket to log this.

ts-test/tests/path_operations.ts
DomPeliniAerospike: Adding to a shared module could reduce the complexity.  Will add ticket for this.

typings/index.d.ts
DomPeliniAerospike: Fixed

typings/index.d.ts
DomPeliniAerospike: fixed

binding.gyp
DomPeliniAerospike: Will add ticket for this.

out.txt
DomPeliniAerospike: Removed.

PR #940, opened 2026-02-09
Node.js Release 6.6.0

DomPeliniAerospike:


.github/workflows/stage-comprehensive-tests.yml
```diff
@@ -1196,7 +1196,7 @@ jobs:
 
     - name: Run latest Community edition Aerospike server
       if: ${{ inputs.run_tests && startsWith(inputs.platform-tag, 'manylinux') }}
-      run: docker run -d --name aerospike -e DEFAULT_TTL=2592000 -p 3000-3002:3000-3002 aerospike/aerospike-server:8.1.0.1
```
juliannguyen4, line 1199: TODO for later: this should take in the `server-tag` input?

examples/pathExpressions.js
```diff
@@ -0,0 +1,54 @@
...
+    floatList: [2.4, 4.8, 7.2]
+  }
+
+  await client.put(key, bins)
+
+  const addAllChildren = new Context().addAllChildren()
+
+  const modExpression = Aerospike.exp.mul(Aerospike.exp.loopVarFloat(Aerospike.exp.loopVarPart.VALUE), Aerospike.exp.float(3.7))
+
+  const modifyByPath = Aerospike.exp.modifyByPath(Aerospike.exp.binList('floatList'), Aerospike.exp.type.LIST, modExpression, Aerospike.exp.pathModifyFlags.DEFAULT, addAllChildren)
```
juliannguyen4, line 41: TODO for later (I understand this PR is urgent): I think having each argument in its own line is easier to read vs. having all of the args in one line

lib/exp.js
```diff
@@ -1101,3 +1272,121 @@ exports.expWriteFlags = {
...
+   */
+  MAP_KEY_VALUE: pathSelectFlags.MAP_KEY_VALUE,
+
+  /**
+   * If the expression in the context hits an invalid type (e.g., selects
+   * as an integer when the value is a string), do not fail the operation;
+   * just ignore those elements. Interpret an expression that returns UNKNOWN
+   * as false instead.
+   *
+   * The list is formatted as [key0, value0, key1, value1,...].
```
juliannguyen4: This should be in the docstring for `MAP_KEY_VALUE`

lib/operations.js
```diff
@@ -192,3 +195,45 @@ exports.touch = function (ttl) {
...
+    flags
+  })
+}
+
+/**
+ * @summary Path modify expression.
+ * Used in path expressions.
+ *
+ * @param {string} bin - The name of the bin.
+ * @param {AerospikeExp} modExp - expression for modificaiton.
```
juliannguyen4: typo
```suggestion
 * @param {AerospikeExp} modExp - expression for modification.
```

src/main/util/conversions.cc
```diff
@@ -58,18 +58,27 @@ extern "C" {
...
 const char *GeoJSONType = "GeoJSON";
+const char *HyperLogLogType = "HyperLogLog";
 const char *BinType = "Bin";
 const char *TransactionType = "Transaction";
 
 const int64_t MIN_SAFE_INTEGER = -1 * (std::pow(2, 53) - 1);
 const int64_t MAX_SAFE_INTEGER = std::pow(2, 53) - 1;
 const uint64_t UMAX_SAFE_INTEGER = std::pow(2, 53) - 1;
+
+bool g_wrap_hll = false;
```
juliannguyen4, line 76: `bool g_wrap_hll = false;`

Just wondering why we need a global flag to convert C client's hyperloglog to Node.js? Python client doesn't use a global bool for this

src/main/util/conversions.cc
```diff
@@ -1378,6 +1398,32 @@ int asval_from_jsvalue(as_val **value, Local<Value> v8value, const LogInfo *log)
 		*value = (as_val *)as_integer_new(int64_value);
 #endif
 	}
+	else if (is_hyperloglog_value(v8value)) {
```
juliannguyen4, line 1401: Long term TODO you could combine this with the other codepath that converts regular Javascript blobs to C client as_bytes

src/main/expressions.cc
```diff
@@ -198,10 +204,34 @@ int convert_entry(Local<Object> entry_obj, as_exp_entry *entry,
...
+				.ToLocalChecked();
+		if (exp_obj->IsArray()) {
+			Local<Array> exp_ary = Local<Array>::Cast(exp_obj);
+			if (compile_expression(exp_ary, &exp, log) != AS_NODE_PARAM_OK) {
+				return AS_NODE_PARAM_ERR;
+			}
+		}
+		else if (exp_obj->IsString()) {
+			Nan::Utf8String exp_b64(exp_obj.As<String>());
+			exp = as_exp_from_base64(*exp_b64);
```
juliannguyen4, line 224: Long term TODO noticed this block of code earlier, so might be helpful to combine into a helper function

src/main/expressions.cc
```diff
@@ -260,13 +290,13 @@ int compile_expression(Local<Array> entries_ary, as_exp **filter_exp,
...
+			free_entries(entries_ary, entries, i, log);
+			return rc;
 		}
 		entry++;
 	}
 	*filter_exp = as_exp_compile(entries, length);
 
-done:
-	free_entries(entries_ary, entries, log);
+	free_entries(entries_ary, entries, entries_ary->Length(), log);
```
juliannguyen4, line 300: Just curious why the new parameter for `entries_ary->Length()` since we can access it from `entries_ary`?

juliannguyen4 (dismissed):
Non-blocking issues since this PR is urgent

.github/workflows/stage-comprehensive-tests.yml
DomPeliniAerospike: Yes, I should standardize this, will add a ticket for this.

examples/pathExpressions.js
DomPeliniAerospike: I agree with this. I will leave for now since these examples will be revamped in the near future.

src/main/util/conversions.cc
DomPeliniAerospike: I can look into this, might be a tad complicated due to HyperLogLog being an object and Blobs being a buffer type, but we could cut down on some duplicate code for sure.

src/main/expressions.cc
DomPeliniAerospike: I have a note to make a ticket add this as a helper function.

src/main/expressions.cc
DomPeliniAerospike:
```
	for (i = 0; i < length; i++) {
		*entry = {};
		Local<Object> entry_obj =
			Nan::Get(entries_ary, i).ToLocalChecked().As<Object>();
		if ((rc = convert_entry(entry_obj, entry, log)) != AS_NODE_PARAM_OK) {
			as_v8_error(log, "Error converting expression entry: %i", i);
			free_entries(entries_ary, entries, i, log);

```			
While parsing entries, if there is an error, we only free entries that have already been written.  To do this, we need to provide the iterator `i`.

lib/operations.js
DomPeliniAerospike: Fixed this typo.

lib/exp.js
DomPeliniAerospike: Good catch, fixed this.

PR #974, opened 2026-03-05
Replace self hosted mac arm64 runners with Github runners for dev workflow. Fix circuit breaker test not compiling

juliannguyen4:
- Replace SMA2 and SMA3 runners with Github-hosted runners. The integration tests are disabled by default anyways, so we just need to build the bindings
- Fix circuit breaker test not compiling

## TODO
- [ ] ~~Stage tests pass~~ will test in dev branch


juliannguyen4:
Dev workflow passes.

PR #978, opened 2026-03-11
6.6.1 release

juliannguyen4:
## Summary of changes

- Fix circuit_breaker.ts bad syntax. Also Dominic combined the maxErrorRate and errorRateWindow test cases into one
- Use the new OIDC credentials for the JFrog database project. The old clients credentials was manually configured and is deprecated

### Dev changes
- Re-enabled dev workflow tests; confirmed with Steve Vives that it is safe to do as long as we have StepSecurity enabled in this repo.
- Use `jf npm` to upload artifacts instead of the `jf rt upload` command. The old way of using `jf rt upload` does not match the structure of an npm repo.
- Move away from embedding the version with dev number in the package.json file; use tags and release bundles to store the dev number instead, and have the package.json store the next expected release version
- Move away from using JFrog as intermediate storage for bindings to using Github artifacts. The removed  `.github/workflows/dev-upload-addons-to-jfrog.yml` is now in `dev-workflow.yml`
- Only run dev workflow manually instead of on every push to a feature branch, since this creates a bump commit and release bundle.

## Stage changes
- Refactored some of the stage workflow jobs. But this is out of the scope for this release 
- Disabled commit job statuses because they were causing workflows to fail. Don't have time to fix now
- package.json: fix valgrind job not working on node.js 25

## TODO (out of scope)
- [ ] This doesn't support deleting older release bundles for the same incoming release 
- [ ] Build info needs to be generated in the same job the build occurred
- [ ] Instead of installing via "jf rt download", move to "npm install" to run stage tests
- [ ] Delete artifacts from Github
- [ ] Move away from Python client feature branch for promoting release bundles


juliannguyen4:
I addressed the Github security warnings but am unable to resolve the comments

justinlee-aerospike (approved):
changes look good

PR #992, opened 2026-03-24
[CLIENT-4003] Add support for bin projections.

juliannguyen4:
- [x] QueryOptions' `ops` attribute is not documented
- [ ]


juliannguyen4:
Without transpiling ts test files to js, only the js files will be run by default

PR #998, opened 2026-04-03
[CLIENT-4435] Add path expressions enhancements only compatible with server >= 8.1.2

sfalvo-aerospike:
Update to the latest stage branch in C client

## Extra Changes
- We rename the typescript tests from `.ts` to `.mts` since the Node.js client uses CommonJS syntax, and we want the typescript compiler to treat these files with ESM syntax instead of the default CommonJS syntax assumed for our project.
- Commented out scan bin projection test cases because they were causing type errors from the typescript compiler. We aren't running them anyways because they are skipped due to a client bug
- Change moduleResolution settings to fix typing errors when running typescript compiler on the tests
- test/tsconfig.json: remove typeRoots since we're only checking the types for the tests. There's errors in node_modules we can't fix
- .mocharc.yml - add test parameters so `npm run gdb/valgrind` will pick these up automatically.

## Notes to keep in mind
- Stage workflow changes may be unfinished which is ok. We will just update in `dev` and then create a PR from `dev` to `stage` once the changes to the workflow are finished.
- Changes to npm gdb command are not tested yet

## Known issues (out of scope and not introduced in this PR)
- [ ] TS type definitions: In other parts of the API, map keys are only allowed as strings and not integers or blobs.

## TODO
- [ ] Move shared workflow for deploying server away from "temporary" commit
- [ ] Run some of the stage tests here to prevent finding issues later down the line that may be expensive
- [ ] For type hinting / intellisense, make sure documentation is accurate for javascript exports


lib/aerospike.d.ts
```diff
@@ -1774,6 +1774,33 @@ export namespace cdt {
          * @return {CdtContext} Updated CDT context, so calls can be chained.
          */
         public addMapValue(value: AerospikeBinValue): CdtContext;
+        /**
+         * Look up one or more keys in a map expression.
+         *
+         * @param keys - List of keys to query in the map context.
+         * @return {CdtContext} Updated CDT context, so calls can be chained.
+         */
+        public addMapKeysIn(keys: AerospikeBinValue[]): CdtContext;
```
juliannguyen4: nit: think this may be slightly misleading since map keys can't be cdt's (i.e maps or lists)

[Supported map key types](https://aerospike.com/docs/develop/data-types/collections/map/#:~:text=map%20keys%20are%20restricted%20to%20the%20following%20scalar%20data%20types%20—%20integer%2C%20string%2C%20and%20blob.)

Maybe creating a type alias for supported map key types would be helpful, then we can use it throughout the client API. I noticed whenever map keys are accepted as parameters, our typescript definitions only supports string map keys

lib/aerospike.d.ts
```diff
@@ -1774,6 +1774,33 @@ export namespace cdt {
...
+         * Restrictions:
+         * - Only one and-filter is allowed per context level. Multiple and-filters
+         *   cannot be chained. To combine multiple conditions, use
+         *   {@link exp.and} with a single call to `addAndFilter()`.
+         * - The preceding context entry must not be an expression type;
+         *   i.e., `addAndFilter()` cannot follow {@link addAllChildren} or
+         *   {@link addAllChildrenWithFilter}.
+         * - The and-filter cannot be the first entry in the context chain.
+         *
+         * Evaluation runs after prior context steps (e.g. map key-list selection);
```
juliannguyen4: nit: may be better to add cross reference to mapKeysIn context

lib/exp.js
```diff
@@ -1167,15 +1210,6 @@ exports.modifyByPath = (bin, valueType, modExp, flags, ctx) => {
   ]
 }
 
-/**
- * Indicates entry deletion for modify_by_path.
- *
- * @return {AerospikeExp} result remove expression.
- */
-exports.resultRemove = () => [
```
juliannguyen4, line 1175: Since this is a breaking change, we might want to defer this until the next major client release. But for the next release we can deprecate `resultRemove` in favor of `removeResult`

src/main/cdt_ctx.cc
```diff
@@ -217,9 +217,14 @@ int get_optional_cdt_context(as_cdt_ctx *context, bool *has_context,
 			as_cdt_ctx_add_map_key_create(context, asValue, (as_map_order) 3);
 			as_v8_detail(log, "Adding Map Value context");
 			break;
+		case (AS_CDT_CTX_MAP_KEYS_IN):
+			asval_from_jsvalue(&asValue, v8value, log);
```
juliannguyen4: I think we should check the return value of `asval_from_jsvalue` since it can fail and return a `AS_NODE_PARAM_ERR`. (the return type could be more clear about that though)

src/main/cdt_ctx.cc
```diff
@@ -237,12 +242,26 @@ int get_optional_cdt_context(as_cdt_ctx *context, bool *has_context,
 				return AS_NODE_PARAM_ERR;
 			}
 			break;
+		case (AS_CDT_CTX_EXP | AS_CDT_CTX_AND):
+			if (v8value->IsArray()) {
```
juliannguyen4: nit: might be better to avoid nested if's for better readability. Like I would write it like this (psuedocode):
```
if v8value->IsArray() is false:
    return error
...
if compile expression fails:
    return error
as_cdt_ctx_add_and_filter()
```

src/main/cdt_ctx.cc
```diff
@@ -348,9 +367,14 @@ as_cdt_ctx* get_cdt_context_heap(int* rc,
 			as_cdt_ctx_add_map_key_create(context, asValue, (as_map_order) 3);
 			as_v8_detail(log, "Adding Map Value context");
 			break;
+		case (AS_CDT_CTX_MAP_KEYS_IN):
+			asval_from_jsvalue(&asValue, v8value, log);
```
juliannguyen4: Return value of `asval_from_jsvalue` should be checked. also if `asValue` isn't an `as_list*` type, should we fail out here?

src/main/cdt_ctx.cc
```diff
@@ -370,8 +394,26 @@ as_cdt_ctx* get_cdt_context_heap(int* rc,
...
+				}
+				as_cdt_ctx_add_and_filter(context, exp);
+				as_v8_detail(log, "Adding AND Filter context With Filter context");
+			}
+			else {
+				as_v8_error(log, "error: value should be an expression, null, or undefined");
+				*rc = AS_NODE_PARAM_OK;
+				return context;
+			}
+			break;
```
juliannguyen4: Think it's better to avoid nested if statements if possible. I wrote another comment with an explanation

juliannguyen4 (commented):
~~Noticed the smoke tests failing on 8.1.2.0, and am currently running them in `dev` with the same server version to create a baseline to compare with.~~ In the meantime I left some comments / suggestions. On `dev` the smoke tests pass with server 8.1.2.0 (except for test noise)

Here's the smoke tests using this feature branch and server 8.1.2.0: https://github.com/aerospike/aerospike-client-nodejs/actions/runs/24523574263 There seems to be a lot of regressions from the path expressions tests -- there's no code changes that seem to cause those, so it might be an issue with the tests not cleaning up

Also for future reference -- it seems like the API docs are generated from the typescript definition. From package.json

```
    "build-docs": "npx typedoc --plugin typedoc-plugin-rename-defaults --entryPointStrategy expand lib/aerospike.d.ts",
```
https://typedoc.org is specifically for TS

juliannguyen4:
Just noticed createRecord() is performing put() operations with the server. Since that's part of the test setup, I think it's better to move it to a `before()` hook, and then we should add an `after()` hook to delete the record.

test/exp_list.mts
```diff
@@ -88,6 +88,43 @@ describe('Aerospike.exp_operations', function () {
...
     })
 
+    describe('inList', function () {
+      helper.skipUnlessVersion('>= 8.1.2', this)
+
+      it('is true when bin string is contained in a literal list (exp.inList case 1)', async function () {
+        const key: Key = await createRecord({ color: 'blue', qty: 5 })
+        const ops: operations.Operation[] = [
+          exp.operations.read(tempBin,
+            exp.inList(exp.binStr('color'), exp.list(['red', 'blue', 'green'])),
```
juliannguyen4, line 116: Might be good to add a test where exp.inList checks if an element is in a list bin

test/exp_map.mts
```diff
@@ -118,6 +118,64 @@ describe('Aerospike.exp_operations', function () {
       })
     })
 
+    describe('map_keys and map_values', function () {
```
juliannguyen4, line 121: I suggest to move `createRecord` to `before()`, as well as adding `after()` hook to clean up the record. (I gave more details in another comment)

test/exp_map.mts
```diff
@@ -118,6 +118,64 @@ describe('Aerospike.exp_operations', function () {
...
+      helper.skipUnlessVersion('>= 8.1.2', this)
+
+      it('returns map keys from bin m (exp_map_keys_values case 1)', async function () {
```
juliannguyen4, line 124: Just curious what `exp_map_keys_values` is? (asking for my own understanding)

test/exp_map.mts
```diff
@@ -118,6 +118,64 @@ describe('Aerospike.exp_operations', function () {
...
+        ]
+        const result: AerospikeRecord = await client.operate(key, ops, {})
+        const list = result.bins![tempBin] as string[]
+        expect(list).to.have.length(2)
+        expect(list).to.have.members(['a', 'b'])
+      })
+
+      it('returns map values from bin m (exp_map_keys_values case 2)', async function () {
+        const key: Key = await createRecord({ m: { a: 1, b: 2 } })
+        await orderByKey('m', key)
```
juliannguyen4, line 140: In the python client I just convert the returned map keys / values from a list to a set, so that the order doesn't matter. (this approach also tests that these features work with unordered maps which are the default in the server) It's not a blocking issue but just something to keep note of

lib/exp.js
juliannguyen4: Currently there's a failing test case related to this

sfalvo-aerospike: How can we be getting such different test results?  Nothing broke for me when I removed this declaration.  This lack of test reproducibility is a major cause for concern.

juliannguyen4: From latest smoke tests using latest server release:
```
  17) Aerospike.exp.selectByPath
       Positive tests
         selectByPath
           arguments
             Path Operations
               Positive tests
                 modifyByPath
                   arguments
                     modExp
                       modifies with result remove expression:
     TypeError: exp.resultRemove is not a function
      at Context.<anonymous> (test/path_operations.ts:199:39)
      at process.processImmediate (node:internal/timers:483:21)
```

test/exp_map.mts
sfalvo-aerospike: The test case function in the C client library.  Looks like AI just pasted the test case name verbatim here.

src/main/cdt_ctx.cc
sfalvo-aerospike: There's no way to avoid a nested if here; the inner if statement depends on exp_ary, which isn't created until after we already know that v8value->IsArray() is true.

src/main/cdt_ctx.cc
sfalvo-aerospike: I was wondering about this myself; this case is a slightly modified clone of another case, so I trusted that type-checking would happen elsewhere.  But I can try to add some error handling here.

src/main/cdt_ctx.cc
sfalvo-aerospike: Scratch previous comment -- I think I see what you mean.  I'll try to flatten this.

juliannguyen4: You can define `exp_ary` after checking if `v8value->IsArray()`

lib/aerospike.d.ts
sfalvo-aerospike: Will be fixed in next push.  Thanks!

lib/exp.js
sfalvo-aerospike: restored

lib/exp.js
```diff
@@ -632,6 +632,57 @@ exports.cmpRegex = (options, regex, cmpStr) => [
...
+/**
+ * Return a remove_result object to indicate entry deletion for modifyByPath.
+ *
+ * @return {AerospikeExp} - nothing
+ */
+exports.removeResult = () => [{ op: exp.ops.REMOVE_RESULT, count: 1 }]
+
+/**
+ * Return a remove_result object to indicate entry deletion for modifyByPath.
+ * This is a convenient alias for {@link removeResult}.
```
juliannguyen4, line 658: Should be marked as deprecated in the docstring

juliannguyen4: Other than that, LGTM

juliannguyen4 (approved):

PR #1003, opened 2026-04-13
[CLIENT-4003] Add support for bin projection. Also transpile and run typescript tests in memory

juliannguyen4:
### CI/CD
- add smoke tests that builds and runs all the dev tests for the node.js client.
- Allow `dev-build-node-cpp-addons.yml` to be run manually so we can test all supported platforms on a feature branch.

### Test changes
- Move `ts-test` folder containing the Typescript tests to `test` folder and the `test` folder containing the Javascript tests to `old-test`. It seems like the Javascript tests haven't been maintained for a while.
- Move test dependencies from `ts-test/package.json` to the root-level `package.json`. I believe it's simpler to have a single `package.json` with all the development-related dependencies
- Compile typescript tests on the fly so that we don't have to save the compiled javascript test files on disk in order to run them.
- In the typescript tests, replace importing the node.js client with relative paths. I believe this is more accurate such that aerospike doesn't need to be installed in the typescript test directory's `node_modules` folder. Also with this approach, we don't need to have a `package.json` for the typescript test folder.

Remove `tsconfig.tsbuildinfo` since this is just for caching the typescript tests when compiling them. 

## Known issues
- [CLIENT-4648](https://aerospike.atlassian.net/browse/CLIENT-4648): Foreground scans are unable to perform bin projection. This is ok for now as long as the Query class is able to do the same thing.
- CLIENT-4650: Circuit breaker test case and other tests can be flakey
- There are some unused imports in the typescript tests

## TODO
- [x] Write tests for scans as well
- [x] Make sure docs can be generated with changes to package.json
- [x] Docs generated from typescript type hints or js source files?

## Notes
- We can make CI/CD changes in `dev` without a PR

[CLIENT-4648]: https://aerospike.atlassian.net/browse/CLIENT-4648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juliannguyen4:
Bin projection tests pass locally with server 8.1.2

sfalvo-aerospike (approved):
Make it so.  </picard>

PR #1004, opened 2026-04-15
[CLIENT-4311] Add support for creating set indices

sfalvo-aerospike:
## Extra changes
- Javascript type hinting - add missing documentation for index_datatype BLOB


src/main/enums/index.cc
```diff
@@ -45,5 +49,6 @@ Local<Object> indexType()
 	set(obj, "LIST", AS_INDEX_TYPE_LIST);
 	set(obj, "MAPKEYS", AS_INDEX_TYPE_MAPKEYS);
 	set(obj, "MAPVALUES", AS_INDEX_TYPE_MAPVALUES);
+	set(obj, "SET", AS_INDEX_TYPE_SET);
```
juliannguyen4, line 52: This should be documented in aerospike.d.ts in order for it to show up in the API docs

test/index.ts
```diff
@@ -156,6 +156,30 @@ context('secondary indexes', function () {
...
     })
+
+    it('should support creating set indexes', function () {
+      const options = {
+        ns: helper.namespace,
+        set: helper.set,
+        index: testIndex.name,
+		bin: "unused",
+		type: Aerospike.indexType.SET,
+		datatype: Aerospike.indexDataType.DEFAULT
```
juliannguyen4, line 167: In `createIndex` it'd be better to document which `indexDataType` to use if using `Aerospike.indexType.SET`

test/index.ts
```diff
@@ -156,6 +156,30 @@ context('secondary indexes', function () {
...
+      }
+
+      return client.createIndex(options)
+        .then((job: IJ) => job.wait(10))
+        .then(() => client.createIndex(options)
+          .catch((error: any) => {
+            if (error.code === Aerospike.status.ERR_INDEX_FOUND ||
```
juliannguyen4, line 174: Shouldn't we assert that an INDEX_FOUND error is raised and not AEROSPIKE_OK?

src/main/enums/index.cc
juliannguyen4: Also should be documented in the javascript source files for non-typescript developers

src/main/enums/index.cc
```diff
@@ -34,6 +34,10 @@ Local<Object> indexDataType()
 	set(obj, "NUMERIC", AS_INDEX_NUMERIC);
 	set(obj, "GEO2DSPHERE", AS_INDEX_GEO2DSPHERE);
 	set(obj, "BLOB", AS_INDEX_BLOB);
+	set(obj, "INTEGER", AS_INDEX_INTEGER);
```
juliannguyen4, line 37: nit: it may be better to delay adding this until the next release

test/index.mts
```diff
@@ -156,6 +156,30 @@ context('secondary indexes', function () {
...
+		bin: "unused",
+		type: Aerospike.indexType.SET,
+		datatype: Aerospike.indexDataType.DEFAULT
+      }
+
+      return client.createIndex(options)
+        .then((job: IJ) => job.wait(10))
+        .then(() => client.createIndex(options)
+          .catch((error: any) => {
+            if (error.code === Aerospike.status.ERR_INDEX_FOUND ||
```
juliannguyen4: Nvm, already found this in the python client (I added a note in the python client repo's commit `a0c6a4ebe56590fdf3a48ac8bbf62d7fbb818a98`)

juliannguyen4 (approved):

src/main/enums/index.cc
juliannguyen4: @sfalvo-aerospike mentioned this was added to be consistent with the other clients' implementation for set indexes support

PR #1010, opened 2026-04-21
6.7.0 release

juliannguyen4:
## Changes in addition to ones made in 8.1.2 feature PR's

- CI/CD: Allow selecting the bump type when bumping the dev number after a release version
- Add back script to delete .node files that aren't relevant to the user's platform. This is a temporary fix until we start using an S3 bucket
- While working on this release, I started noticing the tests were running out of order in mocha. This caused the admin test cases to fail because they depended on the ordering of the test cases. So for admin.ts, I changed the test cases so that they can run independently of each other
- admin.ts: add `skipAll` helper to avoid conflicting with beforeEach fixtures

## TODO

- [x] ~~Script doesn't seem to trigger on test-ee~~
- [x] ~~Metrics client version test is failing; suspect that the wrong JFrog build is being pulled~~
- [x] If no longer using tsx, remove from package.json and lock file
- [x] Confirmed locally that the memory error from metrics.cc also happens on node.js client v6.6.0 using the tests from latest `dev` here and server 8.1.1. This means the memory error wasn't introduced in v6.7.0, but we need to add this error to a suppressions file for now

## Known Issues
- all_children_with_filter context doesn't accept expressions as a base64 encoded string
- Path expression `mapKeysIn` is returning a server error

## Out of scope
- Valgrind maybe could run "#slow" tests if we increase mocha's default timeout from 2 seconds
- CI/CD: generating type definition file from js source code
- [ ] macOS self hosted runners fails to clean up files properly outside a checked out repo


justinlee-aerospike (dismissed):
changes are okay on first review.  Let's have a debrief on all of the tech debt fixes that were needed for this after release.

justinlee-aerospike (approved):
change looks good

PR #1016, opened 2026-04-28
6.7.0 release (#1010)

juliannguyen4:
Add support for server 8.1.2 features:
- Creating set indexes via the client API
- Bin projection
- Path expressions quality of life enhancements

Changes in addition to ones made in 8.1.2 feature PR's

- CI/CD: Allow selecting the bump type when bumping the dev number after a release version Add back script to delete .node files that aren't relevant to the user's platform. This is a temporary fix until we start using an S3 bucket
- While working on this release, I started noticing the tests were running out of order in mocha. This caused the admin test cases to fail because they depended on the ordering of the test cases. So for admin.ts, I changed the test cases so that they can run independently of each other
- admin.ts: add skipAll helper to avoid conflicting with beforeEach fixtures


juliannguyen4:
Changes already approved in upstream PR's

PR #1022, opened 2026-04-30
[CLIENT-4712] Transpile typescript tests using tsx instead of running compiled javascript files from tsc

juliannguyen4:
## Manual testing
- [x] Smoke tests pass except for noise
- [x] In `dev`, the latest smoke tests has a test failure where only the generated `.js` files are included in the stack trace. [Workflow run](https://github.com/aerospike/aerospike-client-nodejs/actions/runs/26338239490/job/77535542948#step:8:2804)
    - [x] Now in this feature branch, failing test cases show a stack trace with the original `.ts` files, which makes test failures easier to debug. [Workflow run](https://github.com/aerospike/aerospike-client-nodejs/actions/runs/27651006020/job/81774730316?pr=1022#step:8:2799)
- [x] Valgrind shows no memory errors (e.g false positives) with this change. There *is* a memory leak but it's not from the node.js client code. [Workflow run](https://github.com/aerospike/aerospike-client-nodejs/actions/runs/27651297027/job/81775674724).


justinlee-aerospike (approved):
changes looks good

PR #1055, opened 2026-07-15
CLIENT-4755: Dev to stage

atripathi-aerospike:
String ops and enhanced errors

- [Fix dynamic config tests for CLIENT-4977 send_key union semantics](https://github.com/aerospike/aerospike-client-nodejs/pull/1055/commits/9cf4db39741c495b7739139e0e64f7eef1bacc90)
- [[CLIENT-4964] Docs: clarify client config "batch_parent_write" and "b…](https://github.com/aerospike/aerospike-client-nodejs/pull/1055/commits/246ebd6fc2f381d45df25b85e508b3e112fb0faa) 
- CLIENT-5015: initial commit for enhanced errors
- 

[CLIENT-4964]: https://aerospike.atlassian.net/browse/CLIENT-4964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juliannguyen4:
Noticed we are missing this 8.1.3 change...

https://aerospike.atlassian.net/browse/CLIENT-4358

juliannguyen4:
The stage workflow is verified to pass except for during the release bundle promotion step - https://github.com/aerospike/aerospike-client-nodejs/actions/runs/29823442388

We promoted the release bundle manually to TEST to circumvent this bug.

atripathi-aerospike:
> Noticed we are missing this 8.1.3 change...
> 
> https://aerospike.atlassian.net/browse/CLIENT-4358

Now this is already implemented and merged on dev. It landed as [CLIENT-4358](https://github.com/aerospike/aerospike-client-nodejs/pull/1056) (PR #1056, commit e745faa2).

[CLIENT-4358]: https://aerospike.atlassian.net/browse/CLIENT-4358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

justinlee-aerospike (dismissed):
changes look good

atripathi-aerospike:
> Noticed we are missing this 8.1.3 change...
> 
> https://aerospike.atlassian.net/browse/CLIENT-4358

This is being addressed.

.github/workflows/reupload-gh-artifacts-as-jfrog-release-bundle.yml
```diff
@@ -72,7 +86,9 @@ jobs:
       uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
       with:
         name: npm-package
-        path: aerospike-*.tgz
+        path: |
+          aerospike-*.tgz
+          packages/prebuild-*/*.tgz
```
juliannguyen4: nit: above it says we are not shipping prebuilds, so I'm wondering why this line is here?

Worst case the prebuilds folder won't exist,  will just get ignored since "if-no-files-found" default value is to warn.

.github/workflows/stage-comprehensive-tests.yml
```diff
@@ -256,60 +281,45 @@ jobs:
...
-      if: ${{ inputs.platform-tag == 'linux_aarch64' }}
-      run: |
-        if docker ps -aq | grep -q .; then
-          docker rm -f $(docker ps -aq) || echo "Failed to remove one or more containers."
-        else
-          echo "No containers to remove."
-        fi
+    - name: Run CE Aerospike server for package manager smoke test
+      if: ${{ inputs.run_tests && (startsWith(inputs.platform-tag, 'linux_') || inputs.platform-tag == 'darwin_x86_64') }}
+      run: docker run --ulimit nofile=15000:15000 -d --name aerospike -e DEFAULT_TTL=2592000 -p 3000-3002:3000-3002 aerospike/aerospike-server:8.1.1.0
```
juliannguyen4: This is using server 8.1.1 but not the latest server 8.2 RC, just FYI

lib/exp_string.js
```diff
@@ -466,33 +524,39 @@ const repeat = (policy, count, bin) => [
 ]
 
 /**
- * Matches `as_exp_string_regex_replace` expansion in as_exp.h (policy flags
- * uint after the quoted pair). `flags` is part of the public signature for
- * API symmetry with the C macro parameters; wire layout follows the C client.
+ * Matches `as_exp_string_regex_replace` in as_exp.h (CLIENT-4824): quoted pair,
```
juliannguyen4: nit, non-blocking: moving forward I don't think we should reference C client code in the node.js client API docs

.github/workflows/master-build-node-cpp-addons.yml
```diff
@@ -1,8 +1,8 @@
 name: Build Node C++ Addons
 run-name: Build Node C++ Addons (run_tests=${{ inputs.run_tests }}, use-server-rc=${{ inputs.use-server-rc }}, server-tag=${{ inputs.server-tag }})
 
-# Builds manylinux wheels and source distribution
-# Optionally run tests on manylinux wheels
+# Builds native prebuilds
```
juliannguyen4: Important: I think we should get rid of this file, it's caller `.github/workflows/master-push-workflow.yaml`, and the called workflow `master-build-node-cpp-bindings.yml` since we are only building and testing once in the dev workflow. We don't need to rebuild and retest on merge to `master`

test/examples_harness.ts
```diff
@@ -0,0 +1,43 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// *****************************************************************************
+
+'use strict'
+
+/* eslint-env mocha */
+/* global expect */
+
+import { expect } from 'chai'
```
juliannguyen4: nit, nonblocking: I don't think this is really an e2e test that tests the node.js client

src/main/string_operations.cc
```diff
@@ -404,13 +406,30 @@ int add_string_op(as_operations *ops, uint32_t opcode, Local<Object> op,
...
+		if (get_optional_int64_property(&end, &end_defined, op, "end", log) !=
+			AS_NODE_PARAM_OK) {
+			break;
+		}
+		if (end_defined) {
+			ok = as_operations_string_snip(ops, bin, ctx_ptr, pol_ptr, start,
+										   end);
+		}
+		else {
+			ok = as_operations_string_snip_start(ops, bin, ctx_ptr, pol_ptr,
```
juliannguyen4: Just curious why there's a duplicate call to `as_operations_string_snip_start` here?

src/main/enums/subcode.cc
```diff
@@ -0,0 +1,126 @@
...
+
+#define set(__obj, __name, __value)                                            \
+	Nan::Set(__obj, Nan::New(__name).ToLocalChecked(), Nan::New(__value))
+
+Local<Object> subcode()
+{
+	Nan::EscapableHandleScope scope;
+	Local<Object> obj = Nan::New<Object>();
+
+	set(obj, "AEROSPIKE_ERROR_DETAIL_NONE", AS_ERROR_DETAIL_NONE);
```
juliannguyen4: Nonblocking, long term TODO: I would use a macro to define these suberror code constants to reduce code duplication

juliannguyen4: Nonblocking, long term TODO: I would use a macro to define these suberror code constants to reduce code duplication. It makes it easier to validate that the Node.js client's constant name matches the C client constant name

examples/doc/secondaryIndex.js
```diff
@@ -15,18 +14,21 @@
...
 
-async function append (client, argv) {
-  const key = new Aerospike.Key(argv.namespace, argv.set, argv.key)
-  const bins = shared.cli.parseBins(argv.bins)
-  await client.append(key, bins)
-  console.info('Record updated successfully:', key)
+async function runExample ({ client, args, console }) {
+  const job = await client.createIndex({
+    ns: args.namespace,
+    set: args.set,
```
juliannguyen4: I think that "args" should've been removed since namespace and set are already defined in the context as standalone variables.

I used Cursor to confirm this:
```
Yes. args is leftover harness wiring; this example should use the injected ns / set like the others.

The harness already puts those on the context:


context.js
Ln 72–77
const context = {
  client,
  args,
  console,
  ns: args.namespace,
  set: args.set,
And the documented example shape is { client, ns, set, ... }.

queryEqual.js, queryInteger.js, put.js, and exists.js all do that. secondaryIndex.js is the odd one still doing args.namespace / args.set.

```

examples/harness/fixtures/support.js
```diff
@@ -0,0 +1,165 @@
...
+
+async function deleteKeys (client, args, userKeys) {
+  const writePolicy = { ...client.config.policies.write }
+  for (const userKey of userKeys) {
+    try {
+      await client.remove(key(args, userKey), writePolicy)
+    } catch (err) {
+      if (err.code !== Aerospike.status.ERR_RECORD_NOT_FOUND) {
+        throw err
+      }
```
juliannguyen4: long term todo, nonblocking: can use batch_remove instead of single key-value remove calls

examples/harness/fixtures/support.js
```diff
@@ -0,0 +1,165 @@
...
+}
+
+async function assertBin (client, args, userKey, binName, expected) {
+  const readPolicy = { ...client.config.policies.read }
+  const record = await client.select(key(args, userKey), [binName], readPolicy)
+  if (!record || record.bins[binName] !== expected) {
+    const received = record ? record.bins[binName] : undefined
+    throw new Error(`Expected ${binName}=${JSON.stringify(expected)}, got ${JSON.stringify(received)}`)
+  }
+}
```
juliannguyen4: nonblocking note: I don't believe asserting the expected values after running the examples was in scope, but it's not a big deal and we can clarify with Shannon and the rest of the client team later

It's just more code we have to maintain and we already have functional e2e tests that do the same thing

examples/harness/fixtures/validation.js
```diff
@@ -0,0 +1,423 @@
...
+  if (record.bins.bin1 !== 'value1') {
+    throw new Error(`PutGet verification failed: expected bin1=value1, got ${JSON.stringify(record.bins.bin1)}.`)
+  }
+  const header = await client.get(support.key(args, 'putgetkey'), readPolicy)
+  if (!header || !header.gen) {
+    throw new Error('PutGet verification failed: record header was not populated.')
+  }
+}
+
+async function exists (client, args) {
```
juliannguyen4: nonblocking: function name seems to contradict support.assertNotExists

examples/harness/results/junit.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const tests = results.length
+  const failures = results.filter(r => r.status === ExampleResult.Failed).length
+  const skipped = results.filter(r => r.status === ExampleResult.Skipped).length
+  const timeSec = (results.reduce((sum, r) => sum + r.durationMs, 0) / 1000).toFixed(3)
+
+  const cases = results.map(r => {
+    const duration = (r.durationMs / 1000).toFixed(3)
+    if (r.status === ExampleResult.Failed) {
+      return `    <testcase classname="AerospikeExample" name="${escapeXml(r.name)}" time="${duration}">\n` +
```
juliannguyen4: nonblocking: I would use a third party library to generate this junit xml file so we don't have to embed xml strings in javascript code

juliannguyen4 (changes requested):
I've reviewed the PR except for the scripts in `scripts/ci`. The main thing that stands out to me is we are still referring to the node addons as "prebuilds" even though we aren't relying on the "prebuild" npm package anymore. I would rename "prebuilds" to "addons" to make it less confusing for readers

There seems to be a lot of validation steps added to `scripts/ci` but I'm not sure if all of them are necessary..

CHANGELOG.md
```diff
@@ -8,15 +8,23 @@ Breaking changes below require a **major** version bump; they are intended to sh
...
 * **Breaking Changes**
   * `exp.string.concat(policy, second, bin)`: the second argument must be a **string literal** (C `as_exp_string_concat`). List-valued operands must use **`exp.string.concatList(policy, valuesExp, bin)`** (C `as_exp_string_concat_list`).
-  * `exp.string.regexReplace` now takes **`policy` first**, matching the C macro signature; the `flags` argument is reserved (current C expansion does not pack regex flags on the wire).
+  * `exp.string.regexReplace` takes **`policy` first** for API symmetry with other string modify expressions; **regex flags** and **string policy flags** are both packed on the wire (C client `stage`, CLIENT-4824).
   * `strings.snip(bin, start, end)` now requires the **exclusive end** index (half-open range); the one-argument `snip(bin, start)` form is removed.
   * `strings.substrRange(bin, start, end)` — the third parameter is the **exclusive end** index (not a length).
+  * [AER-6957] CVE-2026-63662: C client 7.6.1 **bounds-checks msgpack deserialization** and **rejects nested list/map (CDT) values deeper than 64** (`MSGPACK_MAX_DEPTH`). This matches the server stored-value bound. Reads of deeper nesting now fail instead of unpacking.
```
juliannguyen4:
```suggestion
  * [AER-6957] CVE-2026-63662: C client 7.6.1 **bounds-checks msgpack deserialization** and **rejects nested list/map (CDT) values deeper than 64** (`MSGPACK_MAX_DEPTH`). This matches the server stored-value bound. Reads and writes of deeper nesting now fail instead of unpacking.
```

lib/exp_string.js
atripathi-aerospike: @juliannguyen4 thank you for suggestion . This is done as suggested. Public JSDoc in lib/exp_string.js no longer names C macros or as_exp.h. Kept C references only in the file-header wire-layout notes, not in the published API docs.

.github/workflows/stage-comprehensive-tests.yml
atripathi-aerospike: @juliannguyen4 Replaced the hardcoded 8.1.1.0 image with $CE_SERVER_IMAGE everywhere in comprehensive tests. Stage workflow_dispatch takes server-tag and use-server-rc. Default is public Hub GA 8.1.2.5 so a normal Stage run does not 404. For 8.2 RC set server-tag to 8.2.0.0 (or the TEST timestamp tag) and enable use-server-rc; that pulls from database-docker-virtual, same as Python.

src/main/string_operations.cc
atripathi-aerospike: Non Blocking change and we will fix after node 7.0 release. Just kind of nit. Tracking ticket. https://aerospike.atlassian.net/browse/CLIENT-5532

src/main/enums/subcode.cc
atripathi-aerospike: @juliannguyen4  agreed upon this one, good long-term cleanup. Each set() today repeats the C token as a JS string (AEROSPIKE_SUB_* vs AS_SUB_*), so a typo would not fail the compile. We’ll add a stringify macro after the 7.0 release so the native name is derived from the C identifier. Public Aerospike.subcode.* short names stay the same. Logged as a follow-up; not blocking this release.

examples/doc/secondaryIndex.js
atripathi-aerospike: @juliannguyen4 updated as suggested. Thank you!!!

examples/harness/fixtures/support.js
atripathi-aerospike: @juliannguyen4 done as suggested in the example harness only. Fixture cleanup (deleteKeys / deleteRange / deleteIntegerRange) now uses batchRemove and ignores ERR_RECORD_NOT_FOUND per result. Doc examples still use single-key remove so they match the documented API.

examples/harness/fixtures/support.js
atripathi-aerospike: @juliannguyen4 Agreed it is extra maintenance, and mocha already covers the same client behavior. Leaving the post-example asserts in for this release so examples:all still fails if a doc snippet writes the wrong bins. We can decide with Shannon after release whether the harness should only setup/cleanup. Logged as a follow-up; not changing it in this PR.

examples/harness/fixtures/validation.js
atripathi-aerospike: @juliannguyen4 The check is correct: the Exists snippet deletes existskey at the end, so the harness asserts not-exists. The function name was the confusing part. Renamed it to existsValidate to match putValidate / deleteValidate.

examples/harness/results/junit.js
atripathi-aerospike: @juliannguyen4 Agreed it would be cleaner long term. Leaving the small hand-rolled writer for this release — it already feeds smoke and Stage JUnit. We’ll pick a library after 7.0 so we can add a dep and confirm GitHub still parses the report. Logged as a follow-up; not blocking.

CHANGELOG.md
atripathi-aerospike: @juliannguyen4 taken this into consideration in the PR. Thank you for pointing.

juliannguyen4 (approved):
I spoke with Abhishek and we aligned with the idea that the "prebuild" naming convention in `scripts/` was kept to be consistent with the previous scripts in that folder. That way it is easier to see the changes before and after. The old scripts are still in this folder and can be restored in case. After this release, I would like to simplify the scripts so it's easier to manually review

PR #1057, opened 2026-07-16
CLIENT-4114: Updated the scripts for stage failures

atripathi-aerospike:
Prebuilds were written as aerospike.115.node, but node-gyp-build expects prebuildify-style names like node.abi115.node. With the old names, every Node version loaded the Node 20 binary, so stage CI only passed on Node 20.

Updates relocate-prebuilds.js and CI verify steps to use node.abi${modules}.node.


juliannguyen4 (approved):
Reviewed smoke tests and they pass except for flaky tests

PR #1064, opened 2026-07-19
Fix stage yarn/pnpm install and single-file mocha runs.

atripathi-aerospike:
Use --ignore-scripts when installing the tarball via yarn/pnpm so CI does not rebuild from binding.gyp, and add test-ci-single to run one test file without .mocharc.yml loading the full suite.


dmishra-aero (approved):
Look's good

PR #1065, opened 2026-07-20
Convert remaining legacy test/*.js files to TypeScript.

atripathi-aerospike:
Fix test-ce by passing host/port via OPTIONS instead of mocha CLI flags that conflict with --help.

### DIFF
--- a/.github/workflows/stage-comprehensive-tests.yml
+++ b/.github/workflows/stage-comprehensive-tests.yml
@@ -199,7 +199,7 @@ jobs:
       run: |
         cat .mocharc.yml
         node scripts/verify-prebuild-smoke.js
-        PREBUILDS_ONLY=1 OPTIONS="-t 40000 -h 127.0.0.1 --port 3000 -U superuser -P superuser" npm run test-ci
+        PREBUILDS_ONLY=1 OPTIONS="-t 40000 -h 127.0.0.1 --port 3000 -U superuser -P superuser" npm test --ignore-scripts
 
     - if: ${{ always() }}
       run: docker logs aerospike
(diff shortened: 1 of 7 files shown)


.github/workflows/stage-comprehensive-tests.yml
```diff
@@ -836,7 +836,7 @@ jobs:
       if: ${{ inputs.run_tests && startsWith(inputs.platform-tag, 'linux_') }}
       run: |
         node scripts/verify-prebuild-smoke.js
-        PREBUILDS_ONLY=1 npm run test-ci -- --h localhost --port 3000;
+        PREBUILDS_ONLY=1 OPTIONS="--h localhost --port 3000" npm run test-ci;
```
juliannguyen4: Just curious why we have two commands `test-ci` and `test` in package.json that are identical?

```
    "test": "mocha test/${npm_config_testfile:-}",
    "test-ci": "mocha test/${npm_config_testfile:-}",
```

atripathi-aerospike: test-ci exists so CI can run mocha without triggering posttest → npm run lint. Stage jobs use npm run test-ci for that reason; locally npm test still runs lint afterward. We could use npm test --ignore-scripts in workflows instead, but a dedicated script makes the intent explicit. Happy to add a one-line comment in package.json or consolidate if you prefer.

juliannguyen4: Talked with @atripathi-aerospike and I requested to have a comment in package.json describing this

juliannguyen4 (approved):

PR #1079, opened 2026-08-06
CLIENT-5213: slim prebuild packages and regexReplace flags fix

atripathi-aerospike:
Bump C client to stage (CLIENT-4955) and pack regex flags on the wire for exp.string.regexReplace. Ship native binaries as optional @aerospike/prebuild-* packages so the main tarball stays slim without post-install scripts.


juliannguyen4 (changes requested):
Hey Abhishek, I reviewed the code and PRD for switching to optional dependencies to pull a single compiled addon package: https://aerospike.atlassian.net/wiki/spaces/CLIENTS/pages/5362188295/NodeJs+CI+Refactor

I believe for pattern A, we were supposed to remove node-gyp-build since npm's optionalDependencies should automatically pull the package relevant for the user's platform.

From the PRD:

> Phase 2 (when there is appetite): migrate Pattern B → Pattern A. Pattern B's prebuilds/-/ layout is structurally identical to what Pattern A's sub-packages need. The migration becomes: extract each prebuilds/ subdirectory into its own @aerospike/ package, publish those alongside the main one, change the loader from node-gyp-build to platform-keyed resolution, and list sub-packages as optionalDependencies. No throwaway work.


Otherwise, the packaging looks correct; I'm assuming the [deploy artifacts](https://github.com/aerospike/shared-workflows/blob/main/.github/workflows/deploy-artifacts/README.md) workflow would upload each artifact (*.tgz) as a separate package.

atripathi-aerospike:
> Hey Abhishek, I reviewed the code and PRD for switching to optional dependencies to pull a single compiled addon package: https://aerospike.atlassian.net/wiki/spaces/CLIENTS/pages/5362188295/NodeJs+CI+Refactor
> 
> I believe for pattern A, we were supposed to remove node-gyp-build since npm's optionalDependencies should automatically pull the package relevant for the user's platform.
> 
> From the PRD:
> 
> > Phase 2 (when there is appetite): migrate Pattern B → Pattern A. Pattern B's prebuilds/-/ layout is structurally identical to what Pattern A's sub-packages need. The migration becomes: extract each prebuilds/ subdirectory into its own @aerospike/ package, publish those alongside the main one, change the loader from node-gyp-build to platform-keyed resolution, and list sub-packages as optionalDependencies. No throwaway work.
> 
> Otherwise, the packaging looks correct; I'm assuming the [deploy artifacts](https://github.com/aerospike/shared-workflows/blob/main/.github/workflows/deploy-artifacts/README.md) workflow would upload each artifact (*.tgz) as a separate package.

Address the changes as suggested.

shannonklaus (approved):

PR #1081, opened 2026-08-11
Fix stage comprehensive tests for slim prebuild layout.

atripathi-aerospike:
Stage jobs assumed prebuilds lived inside the main aerospike tarball. Fetch main and platform prebuild packages from JFrog instead.


dmishra-aero (approved):
Look's good

PR #1082, opened 2026-08-11
Fix stage JFrog fetch for slim prebuild release bundles.

atripathi-aerospike:
Download main and platform prebuild packages from the dev release bundle (tag version) with npm path and registry fallbacks, instead of relying on flat scoped npm paths that Stage could not resolve.


dmishra-aero (approved):
changes look's good

PR #1083, opened 2026-08-11
Fix stage JFrog slim package fetch (bundle paths + npm auth).

atripathi-aerospike:
Use bundle-scoped rt downloads and full bundle fallback, configure npm via jf npm config for registry installs, pick the latest 7.0.0-dev.* tag, and set JF_PROJECT=database on all stage JFrog steps.


.github/workflows/stage-comprehensive-tests.yml
```diff
@@ -93,6 +93,7 @@ jobs:
       uses: step-security/setup-jfrog-cli@a6b41f8338bea0983ddff6bd4ede7d2dcd81e1fa # v4.8.1
       env:
         JF_URL: https://aerospike.jfrog.io
```
dmishra-aero, line 95: could use VAR for this `JF_URL, oidc provider name, oid-audience, JF_project`

PR #1095, opened 2026-09-03
[CLIENT-5375] Bind C stage string ops, list join, and bit b64Encode.

atripathi-aerospike:
## Summary

Bind the Node.js client to C `stage` (`06288ec5`) so string, list-join, bit, and error-detail APIs match the 8.1.3 string-ops spec. Nested string CTX, `snipStart`, `CREATE_ONLY`, `lists.join`, `bits.b64Encode`, FLOAT `isNumericType`, canonical matching, the string result-size cap, and `errorDetailVerbosity.EXP_TRACE` are covered on operate and expression paths.

Target server: **Enterprise 8.1.3.0-105**. Nested CTX, `regexReplace` GLOBAL, `b64Encode`, and the result-size PARAM path require that build (not 8.1.3.0-55).

## Tickets

- [CLIENT-5375](https://aerospike.atlassian.net/browse/CLIENT-5375) — Nested string CTX envelope `[0xFF, ctx, [sub_op, args…]]` via C `as_cdt_pack_ctx`. JSDoc on `strings.withContext`. Live tests: list index, map key, 2-entry path. `toString` has no CTX. Expression-path CTX on `exp.string` is out of scope (match Python/C).
- [CLIENT-5359](https://aerospike.atlassian.net/browse/CLIENT-5359) — `strings.snipStart(bin, start)` and 1-arg `strings.snip(bin, start)` pack `[53, start]` only (opcode 53, not 51). `exp.string.snipStart` packs no policy uint.
- [CLIENT-5367](https://aerospike.atlassian.net/browse/CLIENT-5367) — `strings.writeFlags.CREATE_ONLY = 1`. Mutually exclusive with `UPDATE_ONLY`. Invalid with CTX. Matrix: missing bin creates, live bin `BIN_EXISTS`, `NO_FAIL` no-op, combined flags PARAM.
- [CLIENT-5337](https://aerospike.atlassian.net/browse/CLIENT-5337) — `lists.join` / `lists.joinSeparator` (CDT op 28) and `exp.lists.join` / `joinSeparator`. Inverse of `strings.split` / `splitSeparator`. Nested list via `withContext`.
- [CLIENT-5346](https://aerospike.atlassian.net/browse/CLIENT-5346) — `bits.b64Encode` / `b64EncodeFrom` / `b64EncodeRange(..., invertSize)` (BITS op 55) and `exp.bit.*`. No `-1` sentinel; `invertSize` counts size from the end. Inverse of `strings.b64Decode`.
- [CLIENT-5273](https://aerospike.atlassian.net/browse/CLIENT-5273) — FLOAT `isNumericType` docs: requires `.` + a digit (`3.14` yes; `5` / `5.` / `1e5` no). Not `strtod`. Operate + exp truth table.
- [CLIENT-5382](https://aerospike.atlassian.net/browse/CLIENT-5382) — Canonical NFC/NFD `replace`, `startsWith`, `endsWith`. Result-size cap: oversized `padStart`/`repeat` returns PARAM (4), not record-too-big (13). Default Docker `max-record-size` is 1 MiB, so mid-size pads may hit 13 first; 8 MiB hits the string cap.
- [CLIENT-5258](https://aerospike.atlassian.net/browse/CLIENT-5258) — `errorDetailVerbosity.EXP_TRACE = 3`. String subcodes: CTX malformed (8), conversion failed (10), UTF-8 invalid (11), regex limit (12), b64 invalid (13).
- [CLIENT-4955](https://aerospike.atlassian.net/browse/CLIENT-4955) — `exp.string.regexReplace` GLOBAL replaces all matches on 8.1.3.0-105; CASE_INSENSITIVE; NO_FAIL + invalid pattern returns the source string.
- [CLIENT-4824](https://aerospike.atlassian.net/browse/CLIENT-4824) — `exp.string.regexReplace` packs regex flags then policy flags (wire test).
- [CLIENT-5275](https://aerospike.atlassian.net/browse/CLIENT-5275) — C submodule bump to `stage` @ `06288ec5` (C string-ops landing). Node parent records the SHA only.

Parent / reference only (not implemented here): [CLIENT-5330](https://aerospike.atlassian.net/browse/CLIENT-5330), [CLIENT-5331](https://aerospike.atlassian.net/browse/CLIENT-5331), [CLIENT-5376](https://aerospike.atlassian.net/browse/CLIENT-5376).

## Out of scope

- EE string masking
- Stored-bin invalid UTF-8 suite (no raw STRING particle escape hatch)
- Inventing `exp.string` CTX receivers (nested leaves use list/map get expressions, then string exp)

## Test plan

- [x] `OPTIONS="--host 127.0.0.1 --port 3105"` mocha: `strings_operate`, `strings_result_size`, `lists_join`, `bitwise_b64`, `error_detail`, `exp_string_wire`, string/list/bit expression suites
- [x] Nested CTX, regexReplace GLOBAL, b64Encode, snipStart, CREATE_ONLY, join, FLOAT, canonical matching, EXP_TRACE — pass on EE 8.1.3.0-105
- [x] Result-size: skip record-too-big from ns 1 MiB cap, then assert PARAM on 8 MiB pad
- [x] Customer lab `node-client-example` `test:e2e` against this checkout + 105 (`AEROSPIKE_CLIENT_ROOT` + `AEROSPIKE_HOSTS=127.0.0.1:3105`)

## Related

- **Branch:** `CLIENT-5375-string-ops` → `dev`
- **C submodule:** `781c00b7` → `06288ec5`
- **Server:** Aerospike Enterprise 8.1.3.0-105

[CLIENT-5375]: https://aerospike.atlassian.net/browse/CLIENT-5375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLIENT-5359]: https://aerospike.atlassian.net/browse/CLIENT-5359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLIENT-5367]: https://aerospike.atlassian.net/browse/CLIENT-5367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/aerospike.d.ts
shannonklaus: The c client is packing flags for regexReplace now and this comment should reflect that

shannonklaus (changes requested):

lib/aerospike.d.ts
```diff
@@ -12860,6 +12860,22 @@ export namespace bitwise {
...
+     * Create bit "b64 encode" from a byte offset through the end of the blob.
+     * @remarks A negative byteOffset counts back from the end. Requires server 8.1.3+.
+     */
+    export function b64EncodeFrom(bin: string, byteOffset: number): BitwiseOperation;
+    /**
+     * Create bit "b64 encode" on a byte span.
+     * @remarks When invertSize is true, byteSize counts back from the blob end,
+     * so 0 means to the end. Requires server 8.1.3+.
+     */
+    export function b64EncodeRange(bin: string, byteOffset: number, byteSize: number, invertSize?: boolean): BitwiseOperation;
```
juliannguyen4, line 12878: nonblocking: in the future, I believe it would simplify the API to have default values for `byteOffset` (0), `byteSize` (e.g `None` to signal to select through the end of the bytearray), and `invertSize` (`false`). Then you can just have one API call instead of three

This also applies to:
- joinSeparator

lib/aerospike.d.ts
```diff
@@ -12960,6 +12983,10 @@ export namespace strings {
     export enum numericType {
         ANY = 0,
         INT = 1,
+        /**
+         * Floating-point only. Requires `.` plus a digit (not `strtod`):
```
juliannguyen4: nonblocking: I think `strtod` might be confusing for Javascript users who don't write C code

juliannguyen4 (commented):
Besides the issue that Shannon mentioned, LGTM

lib/aerospike.d.ts
atripathi-aerospike: Updated in 0ea95425. The .d.ts comment now matches C stage / CLIENT-4824: exp.string.regexReplace packs regex flags, then string policy flags (NO_FAIL, etc.) on the wire. lib/exp_string.js was already packing both.
Thank you @shannonklaus

lib/aerospike.d.ts
atripathi-aerospike: @juliannguyen4 Addressed in 0ea95425. bits.b64Encode(bin, byteOffset?, byteSize?, invertSize?) now covers whole-blob / from-offset / range (byteOffset omitted = whole blob, byteSize omitted = through the end, invertSize default false). b64EncodeFrom / b64EncodeRange remain as aliases. Same pattern for lists.join(bin, separator?) (joinSeparator is an alias) and the matching exp.bit.b64Encode / exp.lists.join overloads. Thank you for pointing

lib/aerospike.d.ts
atripathi-aerospike: @juliannguyen4 Updated in 0ea95425. FLOAT docs now describe the rule in JavaScript terms (decimal point plus a digit; "3.14" matches, "5" / "5." / "1e5" do not) without referring to strtod. Thank you for pointing.

shannonklaus (approved):

lib/aerospike.d.ts
juliannguyen4: I meant that these three API calls can be combined into one if we make the byteOffset, byteSize, and invertSize parameters optional:
- b64EncodeRange
- b64EncodeFrom
- b64Encode

juliannguyen4: I was thinking we can remove `b64EncodeRange` and `b64EncodeFrom` since these aren't released yet

juliannguyen4: I did more reading about how Typescript projects design optional parameters for API calls..

https://share.google/aimode/A514mHFEf7afMfzNF

This probably would be a better way to design b64Encode with multiple optional params, so that `undefined` doesn't need to be passed to optional params leading up to the one the user wants to set.

juliannguyen4: Apologies, I think this was bad advice; In python, you can pass keyword arguments at any position without having to pass values for preceding parameters. This doesn't seem to be allowed in Typescript

juliannguyen4: Is it possible you can revert this change?

juliannguyen4 (changes requested):
https://github.com/aerospike/aerospike-client-nodejs/pull/1095/changes/b78db458062a79c75ef57a092762be572a021c0f..0ea9542527371dc2c1cfb861c37df10e8f8b8b77#r3936127679

I made a mistake here, sorry. I gave more details in the comment thread

lib/aerospike.d.ts
atripathi-aerospike: @juliannguyen4 Reverted. TypeScript optional positional params would force callers to pass undefined for earlier arguments, so the three APIs stay separate: b64Encode / b64EncodeFrom / b64EncodeRange, and join / joinSeparator. Same on the expression path.

juliannguyen4 (approved):

PR #1100, opened 2026-09-21
Client 4755 stage 8.2 pipeline

atripathi-aerospike:
## Summary
- Stage **workflow_dispatch** now takes `server-tag` and `use-server-rc`. Default is public GA **8.1.2.5** (newest CE/EE on Docker Hub today). 8.2 is not published there yet, so defaulting to `8.2.0.0-rc4` + RC repos would fail docker pull. For Julian’s 8.2 RC run, set `server-tag` (e.g. `8.2.0.0-rc4`) and enable `use-server-rc`.
- Hardcoded `aerospike-server:8.1.1.0` CE runs, EE `use-server-rc: false`, and `docker pull …:latest` all use that input. Addresses the Stage comprehensive-tests 8.1.1 comment on #1055.
- `skipUnlessVersion` skips the suite in `beforeAll`. Capability probes also swallow `ERR_PARAM`, so 8.1.x EE mocha no longer dies in `before` hooks on 8.2-only ops.
- `test-ce` is linux-only. darwin_arm64 never started CE Docker; those jobs were failing in Set up runner / `brew install coreutils`. (`aero-C02GW662Q05F` Set up runner on tsc/prebuild is still infra.)
- Keep uploading `packages/prebuild-*/*.tgz` next to the slim main tarball (Option B). Comment clarifies we are not bundling prebuilds inside `aerospike-*.tgz`.
- Public `exp.string` JSDoc no longer references C client / `as_exp.h`.
- Darwin source builds compile against Node’s `uv.h` and add Homebrew yaml/openssl `-L` paths so Apple Silicon `connect()` does not segfault (Ronen).

## Test plan
- [ ] Open Stage workflow → confirm inputs: `server-tag` default `8.1.2.5`, `use-server-rc` default off, `run_tests` on.
- [ ] Run Stage with defaults: image pull succeeds; 8.2 string/bit suites skip; linux EE mocha does not fail in `before` probes.
- [ ] Run Stage with `server-tag` = local 8.2 RC tag and `use-server-rc` checked: EE/CE use `*-rc` images; 8.2 suites actually run.
- [ ] Confirm darwin_arm64 no longer has `test-ce` jobs.
- [ ] On Apple Silicon: wipe `aerospike-client-c/target` + `build`, `npm run build`, `connect()` without `LDFLAGS` tribal knowledge.
- [ ] Skim `lib/exp_string.js` module docs: no `as_exp.h` / C macro names.


.github/workflows/stage-comprehensive-tests.yml
```diff
@@ -822,7 +856,7 @@ jobs:
       matrix:
         nodejs-tag: ${{ fromJSON(inputs.nodejs-tags) }}
       fail-fast: false
-    if: ${{ inputs.run_tests && (inputs.platform-tag != 'win32_x64' && inputs.platform-tag != 'darwin_x86_64') }}
+    if: ${{ inputs.run_tests && startsWith(inputs.platform-tag, 'linux_') }}
```
juliannguyen4: Here we're excluding the mac arm64 integration tests. Was this intended?

.github/workflows/stage-comprehensive-tests.yml
```diff
@@ -51,6 +56,7 @@ env:
   JF_PROJECT: ${{ vars.JFROG_PROJECT != '' && vars.JFROG_PROJECT || 'database' }}
   JF_OIDC_PROVIDER: ${{ vars.JFROG_OIDC_PROVIDER != '' && vars.JFROG_OIDC_PROVIDER || 'database-gh-aerospike' }}
   JF_OIDC_AUDIENCE: ${{ vars.JFROG_OIDC_AUDIENCE != '' && vars.JFROG_OIDC_AUDIENCE || 'database-gh-aerospike' }}
+  CE_SERVER_IMAGE: aerospike/aerospike-server${{ inputs.use-server-rc && '-rc' || '' }}:${{ inputs.server-tag }}
```
juliannguyen4: I believe we aren't using the `aerospike-server*-rc` Docker repos anymore. In order to fetch RC's, you'd need to use the JFrog endpoint in the `#dev-announcements` Slack channel

scripts/build-commands.sh
```diff
@@ -42,10 +42,24 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
   AEROSPIKE_LIBRARY=${AEROSPIKE_LIB_HOME}/lib/libaerospike.a
   AEROSPIKE_INCLUDE=${AEROSPIKE_LIB_HOME}/include
 
-  LIBUV_DIR=/usr/local/opt/libuv
+  # Node ships libuv. Compiling the C client against Homebrew /usr/local
+  # libuv (Intel keg on Apple Silicon) ABI-mismatches Node and segfaults in
+  # uv_async_init on connect(). Prefer Node's uv.h; fall back to brew --prefix.
```
juliannguyen4: Should homebrew's libuv be skipped entirely in this script if it's compiled for a different CPU architecture? That would prevent the seg fault

test/exp_bit.ts
```diff
@@ -82,8 +82,7 @@ describe('Aerospike.exp_operations', function () {
           ])
           b64EncodeSupported = true
         } catch (error: any) {
-          if (error.code !== Aerospike.status.ERR_OP_NOT_APPLICABLE &&
-              error.code !== Aerospike.status.ERR_REQUEST_INVALID) {
+          if (!helper.isMissingServerOpError(error)) {
```
juliannguyen4: I noticed above that 

`helper.skipUnlessVersion('>= 8.2.0', this)`

...is already called. So for server versions >= 8.2.0, shouldn't these exceptions be considered errors?

```
  export function isMissingServerOpError (error: any): boolean {
    return code === Aerospike.status.ERR_REQUEST_INVALID ||
      code === Aerospike.status.ERR_OP_NOT_APPLICABLE ||
      code === Aerospike.status.ERR_PARAM
...
```

This question also applies to the changes in `test/strings_operate.ts`, and I would check throughout the tests if these errors are being hidden as well when server >= 8.2

juliannguyen4 (commented):
I left some comments. Please take a look

scripts/build-commands.sh
atripathi-aerospike: @juliannguyen4 Yes. scripts/build-commands.sh now skips Homebrew libuv unless lipo -archs includes the host CPU. An Intel keg on Apple Silicon is never used to compile the C client. We still prefer Node’s uv.h first, then native-arch Homebrew only. If neither is present, the script fails instead of falling back to /usr/local.

.github/workflows/stage-comprehensive-tests.yml
atripathi-aerospike: @juliannguyen4 my bad. Not intended as a product skip. Docker and mocha were already linux-only, so darwin_arm64 test-ce only hit wait-for-as-server and went red. Restored the job and run CE docker + mocha on darwin_arm64 as well.

test/exp_bit.ts
atripathi-aerospike: @juliannguyen4 Agreed. Probes now rethrow those codes when the cluster is >= 8.2.0 (throwUnlessMissingOpOnOlderServer). Same change in strings_operate.ts, exp.ts, bitwise_b64.ts, and the string append/prepend probe. Swallowing is only for pre-8.2, where nested before hooks can still run after skipUnlessVersion in beforeAll.

atripathi-aerospike: Resolved as suggested.

.github/workflows/stage-comprehensive-tests.yml
atripathi-aerospike: @juliannguyen4 Docker Hub aerospike-server*-rc is unused, as you said.

Stage now matches Python when use-server-rc is true: OIDC login (database-gh-aerospike provider and audience, same as Python) and pull from artifact.aerospike.io/database-docker-virtual/aerospike-server[-enterprise]:${server-tag}. Default remains public GA 8.1.2.5 on Docker Hub.

For current 8.2.0.0 RC that is server-tag=8.2.0.0 and use-server-rc=true (or a Slack timestamp tag to pin a specific RC).

We kept Node’s existing CE docker run / EE Dockerfile and only changed the image source + login, rather than switching to setup-aerospike-server yet.

If the first RC run 403s on docker pull, we’ll ping #devops. OIDC for npm artifacts already works in this repo; docker pull of database-docker-virtual is the remaining check.

test/test_helper.ts
```diff
@@ -267,6 +267,17 @@ import * as url from "node:url"
       code === Aerospike.status.ERR_PARAM
   }
 
+  /**
+   * Probe catch helper. On servers that already meet minVersion (default 8.2.0),
+   * missing-op status codes are real failures. Older servers may still run
+   * nested `before` hooks after skipUnlessVersion beforeAll, so those stay skippable.
+   */
+  export function throwUnlessMissingOpOnOlderServer (error: any, minVersion = '>= 8.2.0') {
```
juliannguyen4: I don't believe server < 8.2 will ever run this method if it is used in a test suite that skips on server < 8.2. If that is the case, I believe it's better to remove this method entirely

juliannguyen4 (approved):

test/test_helper.ts
atripathi-aerospike: @juliannguyen4 Agreed. skipUnlessVersion already skips the suite in beforeAll on < 8.2, so that helper was dead there and on 8.2 it swallowed real errors. Removed throwUnlessMissingOpOnOlderServer and the probe skips. Suites now rely on skipUnlessVersion('>= 8.2.0') only.

